PR #5, opened 2019-11-15
spec: Update and simplify packaging for Fedora and add EL8 support

Conan-Kudo:
The packaging was a tad bit more complicated than it needed to be for supporting Fedora and openSUSE equally.

This change simplifies it by bringing it up to date with latest packaging practices in Fedora and adding the macro to enable the Python dependency generator for EL8 so that dependencies are correctly auto-generated when building a package for EL8 (with dependencies from EPEL).

Signed-off-by: Neal Gompa <ngompa13@gmail.com>


Conan-Kudo:
@rjfd Note that I don't think EPEL quite provides everything yet for EL8 for this package, but this means you'll be ready for when it does.

sesdev.spec
```diff
@@ -63,24 +63,16 @@ sed -i -e 's/^\s*lv.qemu_use_session = false$//g' seslib/templates/Vagrantfile.j
...
-%else
 %py3_install
-%endif
+%fdupes %{buildroot}%{python3_sitelib}
 
 %files
 %license LICENSE
 %doc CHANGELOG.md README.md
-%{python3_sitelib}/*
+%{python3_sitelib}/ses*
```
smithfarm: why is this needed?

sesdev.spec
```diff
@@ -16,35 +16,35 @@
 #
 
 
+# Enable Python dependency generation on EL8
+%{?python_enable_dependency_generator}
```
smithfarm: do we need to put this inside an `%if` block?

```
%if 0%{?rhel} >= 8
...
%endif
```

sesdev.spec
```diff
@@ -16,35 +16,35 @@
...
+Requires:       python3-click >= 6.7
 Requires:       python3-Jinja2 >= 2.10.1
 Requires:       python3-pycryptodomex >= 3.4.6
 Requires:       python3-PyYAML >= 3.13
 Requires:       python3-setuptools
-%else
-Requires:       python3-jinja2 >= 2.10.1
-Requires:       python3-pycryptodomex >= 3.4.6
-Requires:       python3-pyyaml >= 3.13
 %endif
```
smithfarm: I gather that for CentOS 8 the python dependencies will be generated automagically, but what about for CentOS 7?

sesdev.spec
Conan-Kudo: Fedora Python Packaging Guidelines<sup>[1]</sup> state to not do a universal glob like that. Doing it this way also catches potentially accidents where `%python3_sitelib` is not expanded for some reason and bombs out instead of producing an incorrect package.

[1]: https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_files_to_include

sesdev.spec
Conan-Kudo: There's more work to be done here if you wanted to support CentOS 7. EPEL 7 doesn't have your dependencies packaged, so someone would need to step up and help get those dependencies packaged there first.

sesdev.spec
Conan-Kudo: Not necessary. The `%{?` part of the macro declaration means that if it doesn't exist, it'll just do nothing.

smithfarm: Technically not necessary, perhaps, but I prefer to make things explicit. With the `%if` block I suggested, the comment could be dropped (I also tend to avoid comments in spec files...).

sesdev.spec
smithfarm: I like that it makes the packaging somewhat more explicit, and it also elegantly avoids packaging the `__pycache__` directory that is caught up by the current line.

Could we make it even more explicit by writing:

```
%{python3_sitelib}/sesdev*
%{python3_sitelib}/seslib*
```

?

sesdev.spec
Conan-Kudo: I just realized you're also building for Fedora 29, and it would be useful for that too. Unless you plan to drop F29 support?

sesdev.spec
Conan-Kudo: @smithfarm Have you considered rearchitecting it so that there's no *.py files in the top level `%python3_sitelib`?

Conan-Kudo: @smithfarm I just pushed a change to move sesdev into a package instead of being a module. That should clean things up for you.

sesdev.spec
rjfd: > Unless you plan to drop F29 support?

Negative. We should keep supporting F29

sesdev.spec
rjfd: I don't think we will ever support centos 7

sesdev.spec
smithfarm: `%if 0%{?rhel} >= 8 || 0%{?fedora} >= 29` in that case

sesdev.spec
rjfd: > Have you considered rearchitecting it so that there's no *.py files in the top level %python3_sitelib?

What are the problems of having py files in the top level?

sesdev.spec
Conan-Kudo: It's not needed for Fedora 30 or newer: https://fedoraproject.org/wiki/Changes/EnablingPythonGeneratorsByDefault

sesdev.spec
Conan-Kudo: It means that you wind up with `__pycache__` at the top level too, which is annoying...

sesdev.spec
rjfd: I'm in favor of being more explicit, even if not needed.

Conan-Kudo: Done.

rjfd:
@Conan-Kudo are you building `sesdev` in any other packaging platform like copr or centos koji?

Conan-Kudo:
Planning on building in COPR, why?

Conan-Kudo:
Do you want to set up this project to auto release into COPR?

rjfd:
> Planning on building in COPR, why?

No, I was just wondering if you were.

Conan-Kudo:
@smithfarm You're supposed to package the `__pycache__` directories. Packages that don't are in violation of both Fedora and openSUSE packaging policy.

PR #10, opened 2019-11-16
Add Github Action to publish to OBS

tspmelo:
This action will be triggered on a release creation and
will publish a new package revision in OBS.

Signed-off-by: Tiago Melo <tmelo@suse.com>

### DIFF
--- a/.github/workflows/obs-publish.yml
+++ b/.github/workflows/obs-publish.yml
@@ -1,0 +1,20 @@
+name: Upload OBS Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Publish to OBS
+      uses: tspmelo/publish-obs-action@master
+      with:
+        obs user:  ${{ secrets.OBS_USERNAME }}
+        obs pass:  ${{ secrets.OBS_PASSWORD }}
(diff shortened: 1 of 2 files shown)


.github/workflows/obs-publish.yml
```diff
@@ -0,0 +1,20 @@
...
+    - uses: actions/checkout@v1
+    - name: Publish to OBS
+      uses: tspmelo/publish-obs-action@master
+      with:
+        obs user:  ${{ secrets.OBS_USERNAME }}
+        obs pass:  ${{ secrets.OBS_PASSWORD }}
+        obs email:  'ricardo.dias@suse.com'
+        obs project:  'home:rjdias:sesdev'
+        obs package:  'sesdev'
+        full name:  'Ricardo Dias'
```
rjfd: Can we also use configuration variables to populate these fields?

tspmelo: Yes I will change all of them to use Secrets.
I used it because they are all publicly available, but after giving a second thought, maybe its better to have them as secret so they aren't used in any fork.

smithfarm:
how does this work with `checkin.sh` ?

tspmelo:
I'm using a copy of `checkin.sh` in https://github.com/tspmelo/publish-obs-action and ignore the one that is in the OBS.
The one in OBS was not working with an existing git repo, so I had to make a few changes.

PR #14, opened 2019-11-20
SUMA deployment in "octopus"

rjfd:
Adds the "suma" role, which can be used instead of the "admin" role,
and will install SUMA and adds all other nodes to SUMA.

The SUMA node will have the salt-master, and all other nodes will
have a salt-minion each.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -99,6 +99,7 @@ disable=print-statement,
         too-many-branches,
         too-few-public-methods,
         too-many-instance-attributes,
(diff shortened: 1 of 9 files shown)


rjfd:
To test the SUMA deployment and play with create a deployment with the following command:
```
sesdev create octopus --roles="[suma], [storage], [storage]" --ram=2 suma
```

LenzGr:
Tested and works! Nice work.

PR #16, opened 2019-11-20
seslib: fix 100% cpu usage in non-block stdout reading

rjfd:


rjfd:
@LenzGr this should fix your issue

PR #19, opened 2019-11-28
Implement --repo-priority / --no-repo-priority

smithfarm:
This commit provides a way to turn off the current behavior where sesdev adds
custom repos with an elevated priority.

Also, it drops the setting of elevated priority on repos hard-coded in
VERSION_OS_REPO_MAPPING. These repos are controlled by package maintainers and
are not expected to need elevated priority.

Fixes: https://github.com/rjfd/sesdev/issues/18

### DIFF
--- a/seslib/templates/Vagrantfile.j2
+++ b/seslib/templates/Vagrantfile.j2
@@ -36,7 +36,6 @@ zypper ar {{ os_repo_url }} {{ os_repo_name }}
 {% endfor %}
 
 zypper ar {{ version_repo }} {{ version }}-repo
-zypper mr -p 95 {{ version }}-repo
 
 zypper --gpg-auto-import-keys ref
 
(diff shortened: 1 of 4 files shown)


smithfarm:
@rjfd I have tested this and it works, so it's ready for review.

The tl;dr is "zypper repo priority is evil and should be used only as a workaround to avoid zypper conflicts when the repos are not sane".

I'm not sure if `--repo-priority` is the right name for the option. Maybe `--set-repo-priority` or `--zypper-priority` or something else?

seslib/templates/Vagrantfile.j2
```diff
@@ -36,7 +36,9 @@ zypper ar {{ os_repo_url }} {{ os_repo_name }}
 {% endfor %}
 
 zypper ar {{ version_repo }} {{ version }}-repo
+{% if repo_priority %}
 zypper mr -p 95 {{ version }}-repo
+{% endif %}
```
smithfarm: Another option would be to remove this `zypper mr -p 95` completely. There should not be any reason to elevate the priority of the devel project repo. Then we could go back to elevating priority of custom repos (provided by the user on the command line) by default - i.e. then we could change the `--repo-priority` option to `default=True`.

rjfd:
@smithfarm while you are it, could you also add a section to the README explaining the usage of custom repos? (If you don't have time no worries)

smithfarm:
@rjfd My pleasure. Done.

PR #22, opened 2019-12-02
If less than 3 storage nodes, increase number of disks

smithfarm:
Fixes: https://github.com/rjfd/sesdev/issues/20

### DIFF
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -92,15 +92,18 @@ def _parse_roles(roles):
         if role.startswith('['):
             _node = []
             if role.endswith(']'):
-                role = role[:-1]
-                _node.append(role[1:])
+                role = role[1:-1].strip()
+                _node.append(role)
                 _roles.append(_node)
             else:
-                _node.append(role[1:])
+                role = role[1:].strip()
+                _node.append(role)
         elif role.endswith(']'):
-            _node.append(role[:-1])
+            role = role[:-1].strip()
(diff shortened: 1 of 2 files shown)


rjfd:
Instead of calculating the number of storage nodes while parsing the roles string, how about having a function that calculates the number of storage nodes given the roles list (the result of the roles parsing function)?

That way we can avoid that global variable...

sesdev/__init__.py
```diff
@@ -215,6 +220,14 @@ def create():
     """
 
 
+def _count_storage_nodes(roles):
+    storage_nodes = 0
+    for node in roles:
+        if 'storage' in node:
+            storage_nodes += 1
+    return storage_nodes
```
rjfd: What about?

```python
return len([node for node in roles if 'storage' in n])
```

smithfarm: That's beyond my comprehension level. What does `n` mean in this context?

rjfd: it's a bug :) edited to fix it

rjfd: it creates a list of nodes that have the 'storage' role and then returns the length of the list

smithfarm:
@rjfd Ready to go now (I found it had been breaking the case when the user does not provide either `--roles` or `--single-node` - fixed)

smithfarm:
Oops, this PR introduced a regression, #25, which is fixed by #26

PR #39, opened 2020-01-03
Add option for predefined libvirt networks

jan--f:
This also exposed the domain config option.

Signed-off-by: Jan Fajerski <jfajerski@suse.com>

### DIFF
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -32,7 +32,10 @@ def libvirt_options(func):
                      help='Hostname of the libvirt machine'),
         click.option('--libvirt-user', type=str, default=None,
                      help='Username for connecting to the libvirt machine'),
-        click.option('--libvirt-storage-pool', type=str, default=None, help='Libvirt storage pool'),
+        click.option('--libvirt-storage-pool', type=str, default=None,
+                     help='Libvirt storage pool'),
+        click.option('--libvirt-networks', type=str, default=None,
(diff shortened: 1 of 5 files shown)


sesdev/__init__.py
```diff
@@ -105,6 +108,8 @@ def common_create_options(func):
                      help='SCC organization username'),
         click.option('--scc-pass', type=str, default=None,
                      help='SCC organization password'),
+        click.option('--domain', type=str, default='{}.com',
+                     help='Domain name used'),
```
smithfarm: I'm not sure what `default='{}.com'` does. Would it make sense to add `show_default` so the default value is shown in `--help`?

toabctl: `default={}` does show the default. But adding `show_default` makes that automatically so I'm in favor of adding that (instead of adding the default to every `help` text)

jan--f: hmm not overly familiar with `click`, I just went with what is there already. Maybe changing this would be better in a seperate PR?

As for this particular instance, I'm not sure showing this will be all that helpful. Somewhere this is formatted into an actual domain, not sure what will end up in the help text.

smithfarm:
```
  --libvirt-networks TEXT         Existing libvirt networks to use (single or
                                  comma separated list)
```

smithfarm:
@jan--f Please rebase, and while you're at it, could you change "Domain name used" to "Domain name to use" ?

PR #46, opened 2020-01-18
qa: initial qa integration

smithfarm:
Add --qa-test option that installs and runs the qa tests added to ceph-bootstrap by https://github.com/ceph-bootstrap/pull/33

Also fix annoying transient failure https://github.com/SUSE/sesdev/issues/47

### DIFF
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -100,6 +100,8 @@ def common_create_options(func):
                      help='Custom zypper repo URL. The repo will be added to each node.'),
         click.option('--repo-priority/--no-repo-priority', default=True,
                      help="Automatically set priority on custom zypper repos"),
+        click.option('--qa-test/--no-qa-test', default=False,
+                     help="Automatically run integration tests on the deployed cluster"),
         click.option('--scc-user', type=str, default=None,
                      help='SCC organization username'),
         click.option('--scc-pass', type=str, default=None,
@@ -269,6 +271,7 @@ def _gen_settings_dict(version,
                        ram,
                        disk_size,
                        repo_priority,
+                       qa_test,
                        vagrant_box,
                        scc_user,
                        scc_pass,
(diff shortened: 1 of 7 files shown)


sesdev/__init__.py
```diff
@@ -100,6 +100,8 @@ def common_create_options(func):
                      help='Custom zypper repo URL. The repo will be added to each node.'),
         click.option('--repo-priority/--no-repo-priority', default=True,
                      help="Automatically set priority on custom zypper repos"),
+        click.option('--qa-test/--no-qa-test', default=False,
```
toabctl: Can we keep the testing separated from the `sesdev` code itself? I think it would be useful to be able to call the same testing code also for non-sesdev deployed environments.

seslib/templates/ceph-bootstrap/ceph_bootstrap_deployment.sh.j2
```diff
@@ -85,3 +109,13 @@ exit 0
...
 
 salt -G 'ceph-salt:member' state.apply ceph-salt
+
+{% if qa_test is defined and qa_test is sameas true -%}
+{%- if ceph_bootstrap_git_repo -%}
+/root/ceph-bootstrap/
+{%- else -%}
+/usr/share/ceph-bootstrap/
+{%- endif -%}
+qa/health-ok.sh --total-nodes={{ nodes|length }} --nfs-ganesha-nodes={{ ganesha_nodes }} --igw-nodes={{ igw_nodes }} --mds-nodes={{ mds_nodes }} --mgr-nodes={{ mgr_nodes }} --mon-nodes={{ mon_nodes }} --osd-nodes={{ storage_nodes }} --osds={{ total_osds }} --rgw-nodes={{ rgw_nodes }}
```
sebastian-philipp: https://github.com/SUSE/ceph-bootstrap/pull/33

sesdev/__init__.py
rjfd: This change is not tying the ceph-bootstrap QA tests to sesdev. This change is only adding the possibility to easily run the independent QA tests of ceph-bootstrap in sesdev.

PR #53, opened 2020-01-20
Fix node ip generation

ricardoasmarques:
Fix a regression introduced in https://github.com/SUSE/sesdev/pull/39 where all nodes have the same `*.200` IP address.

### DIFF
--- a/seslib/__init__.py
+++ b/seslib/__init__.py
@@ -491,7 +491,7 @@ class Deployment():
             self.settings.cluster_network = cluster_network
 
     def _generate_nodes(self):
-        node_id = 1
+        node_id = 0
         for node_roles in self.settings.roles:
             if 'suma' in node_roles and self.settings.version not in ['octopus']:
                 raise RoleNotSupported('suma', self.settings.version)
@@ -500,9 +500,9 @@ class Deployment():
                 name = 'admin'
                 fqdn = 'admin.{}'.format(self.settings.domain.format(self.dep_id))
             else:
+                node_id += 1
                 name = 'node{}'.format(node_id)
                 fqdn = 'node{}.{}'.format(node_id, self.settings.domain.format(self.dep_id))
-                node_id += 1
 
             networks = ''
             public_address = None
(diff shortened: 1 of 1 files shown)


ricardoasmarques:
DNM because `sesdev list` is reporting wrong VM names:

```
| Deployments |  Version   |     Status      |                             VMs                              |
-------------------------------------------------------------------------------------------------------------
| octopus        | octopus       | stopped         | admin, node0, node1, node2                                   |
```

instead of

```
| Deployments |  Version   |     Status      |                             VMs                              |
-------------------------------------------------------------------------------------------------------------
| octopus        | octopus       | stopped         | admin, node1, node2, node3                                   |
```

smithfarm:
Hm, do we need the "VMs" column in the table at all?

seslib/__init__.py
```diff
@@ -491,7 +491,7 @@ def _generate_static_networks(self):
             self.settings.cluster_network = cluster_network
 
     def _generate_nodes(self):
-        node_id = 1
+        node_id = 0
```
jan--f:
```suggestion
        node_id = 1
```
Either keep the id at 1 or increment it before it is used. Otherwise you get the naming issue you already marked, as well as using the 200 ip address twice.

Though actually that is also not ideal, as it keeps the loop counter increment somewhere in the loop where its easy to miss.

seslib/__init__.py
```diff
@@ -491,7 +491,7 @@ def _generate_static_networks(self):
...
         for node_roles in self.settings.roles:
```
jan--f:
```suggestion
        for node_id, node_roles in enumerate(self.settings.roles, start=1):
```
This is probably the better way to iterate with an index here. That should solve the list issues and the duplicate ip address.

seslib/__init__.py
ricardoasmarques: Done. I'm now incrementing before using it.

seslib/__init__.py
ricardoasmarques: I think this will not work because we don't want to increment the `node_id` when role is `admin` or `suma`.

ricardoasmarques:
> **DNM** because `sesdev list` is reporting wrong VM names:
> ...
It should be fixed now.

PR #62, opened 2020-01-28
ceph_bootstrap: use "ceph-bootstrap deploy" command to run ceph-salt formula

rjfd:

### DIFF
--- a/seslib/templates/ceph-bootstrap/ceph_bootstrap_deployment.sh.j2
+++ b/seslib/templates/ceph-bootstrap/ceph_bootstrap_deployment.sh.j2
@@ -6,12 +6,11 @@ set -ex
 cd /root
 git clone {{ ceph_bootstrap_git_repo }}
 cd ceph-bootstrap
-zypper -n in autoconf gcc python3-devel python3-pip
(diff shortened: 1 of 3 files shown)


seslib/templates/ceph-bootstrap/ceph_bootstrap_deployment.sh.j2
```diff
@@ -103,4 +103,4 @@ ceph-bootstrap config ls
 exit 0
 {% endif %}
 
-salt -G 'ceph-salt:member' state.apply ceph-salt
+ceph-bootstrap deploy -n
```
smithfarm: can we get a `--ceph-bootstrap-deploy/--no-ceph-bootstrap-deploy` option, defaulting to `True`, and then either use `salt` or `ceph-bootstrap deploy` here depending on the value of that option?

smithfarm: also, could you use the long version of the option `--non-interactive` to make it more explicit?

rjfd: done!

PR #65, opened 2020-01-29
Revamp --num-disks handling

smithfarm:
The old way would silently override some `--num-disk=X` settings. Also, when num_disks was not set explicitly by the user, the tool would set it too low for 1- and 2-node clusters.

### DIFF
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -306,13 +306,20 @@ def _gen_settings_dict(version,
     if os:
         settings_dict['os'] = os
 
-    if num_disks:
-        if storage_nodes and storage_nodes < 3 and num_disks < 3:
-            num_disks = 3
-        settings_dict['num_disks'] = num_disks
+    # if num_disks not given explicitly, the default value of 2 might not be enough
+    # NOTE: num_disks is guaranteed to be initialized by the
+    # "click.option('--num-disks', default=None, ...)" call, above
+    if num_disks is None:
+        if storage_nodes:
+            if storage_nodes == 1:
+                settings_dict['num_disks'] = 4
+            elif storage_nodes == 2:
+                settings_dict['num_disks'] = 3
+            else:
+                # go with the default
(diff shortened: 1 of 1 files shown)


sesdev/__init__.py
```diff
@@ -306,13 +306,23 @@ def _gen_settings_dict(version,
...
 
-    if num_disks:
-        if storage_nodes and storage_nodes < 3 and num_disks < 3:
-            num_disks = 3
-        settings_dict['num_disks'] = num_disks
+    # if num_disks not given explicitly, the default value of 2 might not be enough
+    try:
+        num_disks
+    except NameError:
+        num_disks = None  # num_disks not defined
```
smithfarm: In practice, this `try...except` block might not be needed. It appears that the tool already initializes `num_disk` to `None` when it's not set explicitly on the command line.

So, reviewers, if you don't like it I'm happy to drop it.

kshtsk: Even if it does not initialize, it is better to initialize it, very weird style to check variables with try-catch block

smithfarm: @kshtsk alas, in python we cannot do `if num_disks is not defined`, so this try-catch is what I came up with. What is your recommended idiom for checking if a variable is defined?

toabctl: `getattr` (see https://docs.python.org/3/library/functions.html#getattr)

smithfarm: @toabctl :man_facepalming: of course - I knew about that.

In other news, this comment has already been resolved.

kshtsk: > @kshtsk alas, in python we cannot do `if num_disks is not defined`, so this try-catch is what I came up with. What is your recommended idiom for checking if a variable is defined?

I was trying to say, that we have to make sure we declare num_disk in the code upper, otherwise treat such situations as a bug (uninitialized variable).

PR #68, opened 2020-01-30
ceph_bootstrap_deployment: "ceph-bootstrap -ldebug deploy"

smithfarm:


RESOURCES
```diff
@@ -0,0 +1,4 @@
+Vagrant and Vagrant Boxes for openSUSE and SLE
+(Confluence page created by Dan Čermák)
+https://confluence.suse.com/display/applications/Vagrant+and+Vagrant+Boxes+for+openSUSE+and+SLE
```
rjfd: I don't like to see URLs that are not accessible for everyone in the documentation. Can we transcribe part of that document to this file?

smithfarm: @rjfd Dropped the offending commit.

Issue #58, opened 2020-01-23
sesdev has no way of updating or deleting the Vagrant boxes it uses


smithfarm:
The way sesdev currently works, once a vagrant box for a particular OS target has been downloaded, sesdev will continue to use it forever. The box will eventually go stale and need updating.

To address this issue, implement an `--update-box` option to `sesdev create`

Alternatively, the issue could be addressed by implementing a new subcommand that deletes the vagrant box corresponding to a given target, e.g.: `sesdev delete-box ses7`. Then a subsequent `sesdev create ses7` would download a fresh box.

And `sesdev delete-box --all` could be implemented to enable the user to get rid of all the boxes at once.

smithfarm:
I've heard that deleting a box in vagrant is not so simple, because the box (VM image) also lives inside libvirt's image pool. So `sesdev` will need to first delete the image from libvirt's pool. Only then will vagrant's box removal operation do the right thing.

smithfarm:
# List images stored in libvirt pool

```
sudo virsh vol-list default
```

# Delete image stored in libvirt pool

```
sudo virsh vol-delete --pool default <img_name>
```

PR #69, opened 2020-01-30
Implement "vagrant box list" and "vagrant box remove"

smithfarm:
Fixes: https://github.com/SUSE/sesdev/issues/58

---

TO DO:

- [x] if sesdev executes `vagrant box add` then it should remove the current image in libvirt to guarantee that when `vagrant up` runs, the new box is uploaded to libvirt

### DIFF
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -238,6 +238,84 @@ def list_deps():
 
 
 @cli.group()
+def box():
+    """
+    Commands for manipulating Vagrant Boxes
+    """
+
+
+@box.command(name='list')
+@libvirt_options
+def list_boxes(**kwargs):
+    """
+    List all Vagrant Boxes installed in the system.
+    """
+    click.echo("List of all Vagrant Boxes installed in the system")
+    click.echo("-------------------------------------------------")
+    settings_dict = _gen_box_settings_dict(**kwargs)
+    settings = seslib.Settings(**settings_dict)
+    box_obj = seslib.Box(settings)
+    box_obj.list()
+
+
+@box.command(name='remove')
+@click.argument('box_name')
+@libvirt_options
+@click.option('--force', is_flag=True, callback=_abort_if_false, expose_value=False,
+              help='Allow to remove Vagrant Box without user confirmation',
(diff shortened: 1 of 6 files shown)


seslib/__init__.py
```diff
@@ -344,6 +344,71 @@ def init(cls, working_dir):
...
+        try:
+            tools.run_sync(["sudo", "-n", "true"])
+        except CmdException as error:
+            print("You must have passwordless sudo privileges to run this command.")
+            print("Sorry.")
+            exit(-1)
+        #
+        # verify that the corresponding image exists in libvirt storage
+        removal_candidates = []
+        output = tools.run_sync(["sudo", "virsh", "vol-list", libvirt_storage_pool])
```
rjfd: we need to support the case where the libvirt server is remote. In that case the command would look like this:

```
sudo virsh -c "qemu+ssh://<user>@<ip_address>/session?keyfile=<ssh_priv_key_path>" vol-list <storage-pool>
```
```

Comment (smithfarm) [reply seslib/__init__.py]: ok, I broke down and reimplemented it using the libvirt API Python bindings.

**Comment by smithfarm** [comment]
@rjfd @toabctl Do you know how to get libvirt-devel installed in the tox environments?

File: sesdev/__init__.py [hunk 2]
```diff
@@ -237,6 +237,99 @@ def _status(nodes):
...
+next time "{box_name}" is used.
+""".format(box_name=box_name, image_name=image_to_remove))
+        try:
+            if click.confirm('Do you wish to remove that image now?', default=False):
+                box_obj.remove_image(image_to_remove)
+            else:
+                raise click.Abort()
+        except click.Abort:
+            click.echo()
+            click.echo("Box will be deleted without removing the underlying image")
```
smithfarm: @rjfd I'm tempted to just omit this whole "Do you really, really mean to remove the libvirt image?" dialog. If there is just one image, we can be pretty sure that it's the right one thanks to the stricter name checking I am about to implement :-)

smithfarm: dropped the dialog - the whole thing is much less verbose now, which is hopefully a good thing

PR #74, opened 2020-02-03
pylintrc: ignore libvirt module

smithfarm:
After the introduction of the libvirt dependency, there were reports
that local "tox -elint" started failing with:

    seslib/__init__.py:12:0: E0401: Unable to import 'libvirt' (import-error)

even though libvirt module is installed in the virtualenv.


smithfarm:
I was able to fix the issue using this PR by nuking my virtualenv and installing the `python3-pylint` RPM in my system.

:+1:

PR #77, opened 2020-02-03
spec: add Fedora-specific Requires

smithfarm:

### DIFF
--- a/sesdev.spec
+++ b/sesdev.spec
@@ -38,13 +38,18 @@ BuildRequires:  fdupes
 BuildRequires:  python3-setuptools
 
 %if 0%{?suse_version}
-Requires:       python3-click >= 6.7
 Requires:       python3-Jinja2 >= 2.10.1
 Requires:       python3-libvirt-python >= 5.1.0
-Requires:       python3-pycryptodomex >= 3.4.6
 Requires:       python3-PyYAML >= 3.13
-Requires:       python3-setuptools
 %endif
+%if 0%{?fedora}
+Requires:       python3-jinja2 >= 2.10.1
+Requires:       python3-libvirt >= 5.1.0
+Requires:       python3-pyyaml >= 3.13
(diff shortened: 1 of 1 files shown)


sesdev.spec
```diff
@@ -38,13 +38,18 @@ BuildRequires:  fdupes
...
 Requires:       python3-libvirt-python >= 5.1.0
-Requires:       python3-pycryptodomex >= 3.4.6
 Requires:       python3-PyYAML >= 3.13
-Requires:       python3-setuptools
 %endif
+%if 0%{?fedora}
+Requires:       python3-jinja2 >= 2.10.1
+Requires:       python3-libvirt >= 5.1.0
+Requires:       python3-pyyaml >= 3.13
+%endif
```
smithfarm: Please note: this is on purpose not implemented as an `%else` block. Having a separate conditional block for Fedora is IMO easier to understand and makes it explicit that we are supporting Fedora.

Conan-Kudo:
@smithfarm This change was unnecessary, as [the dependencies were being auto-generated by the Python module dependency generator](https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_automatically_generated_dependencies), which has been [enabled by default since Fedora 30](https://fedoraproject.org/wiki/Changes/EnablingPythonGeneratorsByDefault).

Conan-Kudo:
@smithfarm As a consequence of this change, you now have duplicate dependencies described: https://build.opensuse.org/package/binary/filesystems:ceph/sesdev/Fedora_30/x86_64/sesdev-1.1.5+1582717868.g68df753-11.1.noarch.rpm

smithfarm:
@Conan-Kudo Thanks! After giving the matter some further thought, I prefer to disable the automatic dependency generation. See https://github.com/SUSE/sesdev/pull/140 for the rationale.

Issue #73, opened 2020-02-01
ses5, ses6, ses7 "Devel" repos are not correct


smithfarm:
We need to use the image repos, and not "standard".

PR #83, opened 2020-02-04
seslib: fix Ceph repos for ses5, ses6, ses7

smithfarm:
Fixes: https://github.com/SUSE/sesdev/issues/73

### DIFF
--- a/seslib/__init__.py
+++ b/seslib/__init__.py
@@ -112,7 +112,8 @@ VERSION_PREFERRED_DEPLOYMENT_TOOL = {
 VERSION_OS_REPO_MAPPING = {
     'ses5': {
         'sles-12-sp3': [
-            'http://download.suse.de/ibs/Devel:/Storage:/5.0/SLE12_SP3/'
+            'http://download.suse.de/ibs/Devel:/Storage:/5.0/images/repo/'
+            'SUSE-Enterprise-Storage-5-POOL-x86_64-Media1/'
         ],
     },
     'nautilus': {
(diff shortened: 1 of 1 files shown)


seslib/__init__.py
```diff
@@ -112,7 +112,8 @@ def init(cls, working_dir):
 VERSION_OS_REPO_MAPPING = {
     'ses5': {
         'sles-12-sp3': [
```
holgis: Is 'sles-12-sp3' still the right name?

seslib/__init__.py
```diff
@@ -127,7 +128,8 @@ def init(cls, working_dir):
     },
     'ses6': {
         'sles-15-sp1': [
```
holgis: Is 'sles-15-sp1' still the right name?

seslib/__init__.py
```diff
@@ -146,8 +148,8 @@ def init(cls, working_dir):
     },
     'ses7': {
         'sles-15-sp2': [
```
holgis: Is 'sles-15-sp2' still the right name?

seslib/__init__.py
smithfarm: If the user does not specify the `--os` option, `sesdev create ses5` will use the os called "sles-12-sp3". This is not an official name -- rather, it's just a label that sesdev uses.

seslib/__init__.py
smithfarm: If the user does not specify the `--os` option, `sesdev create ses6` will use the os called "sles-15-sp1". This is not an official name -- rather, it's just a label that sesdev uses.

seslib/__init__.py
smithfarm: If the user does not specify the `--os` option, `sesdev create ses7` will use the os called "sles-15-sp2". This is not an official name -- rather, it's just a label that sesdev uses.

smithfarm:
@holgis This PR is not changing the names of the dict keys. It is only changing the repos applied to the `ses[567]` OS targets.

PR #87, opened 2020-02-04
ceph_bootstrap: restart salt-master after ceph-bootstrap installation

rjfd:
The ceph-salt-formula state files are installed in a custom directory
that is configured in an additional configuration file stored in
/etc/salt/master.d and the salt-master needs to be restarted to load
the state files in that custom directory.


smithfarm:
Should we revert f327795a06f549a9e84e74c97273631a1379306a as well?

rjfd:
> Should we revert [f327795](https://github.com/SUSE/sesdev/commit/f327795a06f549a9e84e74c97273631a1379306a) as well?

I don't think so. I think [f327795] is worth for other deployment versions besides octopus/ses7

PR #89, opened 2020-02-04
Let --debug run "vagrant up/destroy" in debug mode

smithfarm:
With this patch, we get debug-level output from "vagrant up" and
"vagrant destroy", but only when the command fails.
    
When the command succeeds, vagrant sends the debug messages to stderr
and these are ignored by run_async.

Still, this is better than no debug messages at all when vagrant fails.

### DIFF
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -191,10 +191,14 @@ $ sesdev create octopus --roles="[admin, mon, mgr], \\
        --num-disks=4 --disk-size=10 my_octopus_cluster
 
     """
+    if debug:
+        logger.info("Debug mode: ON")
+        seslib.GlobalSettings.DEBUG = debug
+
     if log_file:
         logging.basicConfig(format='%(asctime)s [%(levelname)s] [%(name)s] %(message)s',
                             filename=log_file, filemode='w',
-                            level=logging.INFO if not debug else logging.DEBUG)
+                            level=logging.DEBUG if debug else logging.INFO)
     else:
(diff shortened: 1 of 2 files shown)


sesdev/__init__.py
```diff
@@ -189,11 +189,15 @@ def cli(work_path=None, config_file=None, debug=False, log_file=None):
...
         logging.basicConfig(format='%(asctime)s [%(levelname)s] [%(name)s] %(message)s',
                             filename=log_file, filemode='w',
-                            level=logging.INFO if not debug else logging.DEBUG)
+                            level=logging.DEBUG if debug else logging.INFO)
     else:
         logging.basicConfig(format='%(asctime)s [%(levelname)s] [%(name)s] %(message)s',
                             level=logging.CRITICAL)
 
+    if debug:
+        logger.info("Debug mode: %s", "ON" if debug else "OFF")
```
ricardoasmarques: Shouldn't this `logger.info` be moved to outside of the `if` condition? Otherwise "Debug mode: OFF" will never be printed.

smithfarm: I put it inside the conditional because printing "Debug mode: OFF" on every run seemed undesirable. Does the user need visual confirmation that debug is off?

ricardoasmarques: In that case we should change this line to:

```
logger.info("Debug mode: ON")
```

smithfarm: Change already made on Friday. Resolving the conversation.

PR #90, opened 2020-02-04
[AFTER: #84] seslib: add SES7 Internal Media when --qa-test given

smithfarm:
Without this patch, the command

    sesdev create ses7 --qa-test ses7_test1

fails because it cannot find ceph-bootstrap-qa.

The package ceph-bootstrap-qa is available on the SES7 Internal Media.

### DIFF
--- a/seslib/__init__.py
+++ b/seslib/__init__.py
@@ -12,10 +12,11 @@ from jinja2 import Environment, PackageLoader
 import libvirt
 
 from . import tools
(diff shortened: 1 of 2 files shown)


smithfarm:
This breaks octopus --qa-test: 

```
(venv) $ sesdev create octopus --single-node --qa-test octopus_test1
Combination of version 'octopus' and OS 'leap-15.2' not supported with --qa-test
```

PR #92, opened 2020-02-05
adapted install from source guide

capri1989:

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,10 +83,20 @@ make sure that package is installed:
 ```
 $ sudo zypper -n install libvirt-devel
 ```
+Once libvirt-devel is installed in the system, you can proceed to install the needed
+packages and to clone the sesdev source code repo, create and activate a virtualenv, 
+and install sesdev's Python dependencies:
 
-Once libvirt-devel is installed in the system, you can proceed to clone the
-sesdev source code repo, create and activate a virtualenv, and install sesdev's
-Python dependencies:
+#### Install needed packages and repositories on openSUSE
+
+```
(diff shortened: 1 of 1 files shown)


smithfarm:
I don't know - this repeats information from further up in the document, and it doesn't seem to take into account the possibility that the user might be using a remote libvirtd?

smithfarm:
I looked again, and I think if you start reading here:

https://github.com/SUSE/sesdev#installation

and read until the end, everything should be fine. The problem happens if you start reading here:

https://github.com/SUSE/sesdev#install-sesdev-from-source

Correct? In that case, would it be sufficient to add a sentence to "Install sesdev from source" saying something like:

"The instructions in this section assume that the user has already completed all the previous instructions in this 'Installation' chapter."

?

capri1989:
> I looked again, and I think if you start reading here:
> 
> https://github.com/SUSE/sesdev#installation
> 
> and read until the end, everything should be fine. The problem happens if you start reading here:
> 
> https://github.com/SUSE/sesdev#install-sesdev-from-source
> 
> Correct? In that case, would it be sufficient to add a sentence at the beginning of the "Install sesdev from source" section - something like:
> 
> "The instructions in this section assume that the user has already completed all the previous instructions in this 'Installation' chapter."
> 
> ?

No, as packages are missing that are needed for the source install and usage such like gcc, python-devel etc etc.

Also vagrant libvirt is documented underneath install from packages and not source or in general, so we either have to move it somewhere else or mention it here again

smithfarm:
I have opened  #94 to address the concerns raised here. Feel free to reopen if you don't like that.

PR #93, opened 2020-02-05
libvirt/network: autostart networks per default

jschmid1:
enabled here: https://github.com/vagrant-libvirt/vagrant-libvirt/pull/708/commits/04b34f8c13f824129b1addb1d21c685f0203da75

Will enable 'autostart' in the libvirt network configuration.

Signed-off-by: Joshua Schmid <jschmid@suse.de>

### DIFF
--- a/seslib/__init__.py
+++ b/seslib/__init__.py
@@ -668,11 +668,11 @@ class Deployment():
             else:
                 if 'admin' in node_roles or 'suma' in node_roles:
                     public_address = '{}{}'.format(self.settings.public_network, 200)
-                    networks = ('node.vm.network :private_network, ip:'
+                    networks = ('node.vm.network :private_network, autostart: true, ip:'
                                 '"{}"').format(public_address)
                 else:
                     public_address = '{}{}'.format(self.settings.public_network, 200 + node_id)
-                    networks = ('node.vm.network :private_network, ip:'
+                    networks = ('node.vm.network :private_network, autostart: true, ip:'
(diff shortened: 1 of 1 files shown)


seslib/__init__.py
```diff
@@ -663,16 +663,17 @@ def _generate_nodes(self):
                 for network in self.settings.libvirt_networks.split(','):
                     networks += (
                         'node.vm.network :private_network,'
+                        'autostart: true,'
```
jan--f: Hmm not sure if we should change the config of manually supplied existing networks.

jschmid1: I wouldn't mind. But I'm also not strongly attached to this as long as _newly_ created networks are started automatically.

smithfarm:
I don't understand what undesirable behavior this is intended to address.

jschmid1:
> I don't understand what undesirable behavior this is intended to address.

networks are not started automatically when not flagged with the 'autostart' bit. This is annoying when you reboot your libvirt host and have to start each network before you can start the VMs.

seslib/__init__.py
jan--f: I'd vote to leave this out. The idea is that the user can pass custom, self-managed networks, so we shouldn't change the configuration of them :)
But I won't block this if there're strong feelings about this.

jschmid1: I guess you're right. We shouldn't change the behavior of an existing network. Removed this line.

PR #103, opened 2020-02-12
CaaSP setup

capri1989:
Ricardo and I added the possibility to do a CaaSP deployment with sesdev. 
You can now either just spawn a completely configured CaaSP environment for development purposes or you could also directly add rook+ceph on top of it.

Example:

`sesdev create caasp4 caasp4` (gives you just a CaaSP cluster)

`sesdev create caasp4 caasp4 --deploy-ses` (gives you a CaaSP cluster with already rook+Ceph on top)`

If you want to have multiple workers or master you could work with the `--roles` option as follows:

`--roles="[master],[worker],[worker],[loadbalancer],[storage]"`

### DIFF
--- a/seslib/templates/provision.sh.j2
+++ b/seslib/templates/provision.sh.j2
@@ -20,7 +20,7 @@ zypper ar {{ _repo }} {{ version }}-repo{{ loop.index }}
 zypper mr -p 98 {{ version }}-repo{{ loop.index }}
 {% endfor %}
 
-{% if version == 'ses7' %}
+{% if version == 'ses7' or version == 'caasp4' %}
 zypper ar -f http://download.suse.de/ibs/SUSE:/CA/SLE_15_SP2/SUSE:CA.repo
 {% endif %}
 
@@ -32,7 +32,7 @@ zypper -n install vim git-core iputils jq make iptables patch man
 zypper -n install vim git iputils hostname jq make iptables patch man
 {% endif %}
 
-{% if version == 'ses7' %}
+{% if version == 'ses7' or version == 'caasp4' %}
 zypper -n install ca-certificates-suse
 {% endif %}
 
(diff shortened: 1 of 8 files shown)


seslib/templates/provision.sh.j2
```diff
@@ -20,7 +20,7 @@ zypper ar {{ _repo }} {{ version }}-repo{{ loop.index }}
 zypper mr -p 98 {{ version }}-repo{{ loop.index }}
 {% endfor %}
 
-{% if version == 'ses7' %}
+{% if version == 'ses7' or version == 'caasp4' %}
 zypper ar -f http://download.suse.de/ibs/SUSE:/CA/SLE_15_SP2/SUSE:CA.repo
```
smithfarm: `caasp4` is deploying on `sles-15-sp1`, but here we are putting the `SLE_15_SP2` version of `SUSE:CA.repo` on it.

I realize that it probably works fine, but I'd be happier if we could put this in a separate conditional block. Since it wouldn't hurt to add `SUSE:CA.repo` on `ses6` as well, it could say:

```
{% if version == 'ses6' or version == 'caasp4' %}
zypper ar -f http://download.suse.de/ibs/SUSE:/CA/SLE_15_SP1/SUSE:CA.repo
{% endif %}

File: seslib/templates/provision.sh.j2 [hunk 2]
```diff
@@ -32,7 +32,7 @@ zypper -n install vim git-core iputils jq make iptables patch man
 zypper -n install vim git iputils hostname jq make iptables patch man
 {% endif %}
 
-{% if version == 'ses7' %}
+{% if version == 'ses7' or version == 'caasp4' %}
```
smithfarm: And this would become:

```
{% if version == 'ses7' or version == 'ses6' or version == 'caasp4' %}
```

seslib/templates/provision.sh.j2
capri1989: Good catch, fixed

smithfarm: nevermind - we can fix this in a follow-up PR.

seslib/templates/provision.sh.j2
smithfarm: nevermind - let's leave this for a follow-up PR

smithfarm:
I tested that this does not break either "ses5" or "ses6" deployments.

PR #133, opened 2020-03-03
provision.sh: remove Python 2 so it doesn't pollute the environment

smithfarm:


smithfarm:
```
admin:~ # rpm -q python-base
package python-base is not installed
```

PR #140, opened 2020-03-05
spec: disable RH/Fedora Python dependendency generator

smithfarm:
According to
https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_automatically_generated_dependencies

"Packages MAY use the automatic Python dependency generator."

"This generator is enabled by default in Fedora. If a packager wishes to
explicitly opt out of the generator . . ., a packager SHOULD opt out
explicitly by adding:

%{?python_disable_dependency_generator}

Although this statement can be used anywhere in the spec, we recommend
putting it just before the main package’s %description declaration."

As maintainer of the upstream spec file, I want to opt out, because
I prefer to explicitly declare (and track) the dependencies in the spec
file.

The dependencies defined in `setup.cfg` are used when sesdev is installed using `pip`.

The dependencies defined in `sesdev.spec` are used when sesdev is installed using `rpm`.

Fedora offers an automated dependency generator. openSUSE does not. It's confusing to list Python dependencies only for one, but not the other.

Obviously, the dependencies in the spec file must be kept in sync with those in `setup.cfg`, otherwise it will break. Activating an automated dependency generator which is available only on Fedora does not relieve me of this responsibility, since I still have to do it for openSUSE.

Being explicit about the dependencies adds information value to the spec file, too, because at once glance I can tell you which RPM package meets the dependency on Fedora, whereas `setup.cfg` does not tell me this information.

### DIFF
--- a/sesdev.spec
+++ b/sesdev.spec
@@ -16,10 +16,6 @@
 #
 
 
-%if 0%{?el8} || (0%{?fedora} && 0%{?fedora} < 30)
-%{python_enable_dependency_generator}
-%endif
-
 Name:           sesdev
 Version:        1.1.6
 Release:        1%{?dist}
@@ -42,7 +38,7 @@ Requires:       python3-Jinja2 >= 2.10.1
 Requires:       python3-libvirt-python >= 5.1.0
 Requires:       python3-PyYAML >= 3.13
 %endif
-%if 0%{?fedora}
+%if 0%{?el8} || 0%{?fedora}
 Requires:       python3-jinja2 >= 2.10.1
 Requires:       python3-libvirt >= 5.1.0
 Requires:       python3-pyyaml >= 3.13
@@ -52,6 +48,10 @@ Requires:       python3-pycryptodomex >= 3.4.6
 Requires:       python3-setuptools
 Requires:       vagrant
 Requires:       vagrant-libvirt
+
+%if 0%{?el8} || 0%{?fedora}
+%{?python_disable_dependency_generator}
+%endif
 
 %description
 sesdev is a CLI tool for developers to help with deploying SES clusters.


smithfarm:
@Conan-Kudo JFYI

Conan-Kudo:
You're already tracking them in setuptools files. If you forget to keep them in sync, the program will fail to load. In general, I do not suggest doing this.

smithfarm:
@Conan-Kudo The setuptools files are not relevant for RPM packages.

Conan-Kudo:
They are, you are literally using them to configure and install the program in the packaging. Python validates the dependencies at the startup of the program, so if you forget, it'll break.

smithfarm:
@Conan-Kudo I am familiar with how Python dependencies work :-)

smithfarm:
The dependencies defined in `setup.cfg` are used when sesdev is installed using `pip`.

The dependencies defined in `sesdev.spec` are used when sesdev is installed using `rpm`.

Fedora offers an automated dependency generator. openSUSE does not. It's confusing to list Python dependencies only for one, but not the other.

Obviously, the dependencies in the spec file must be kept in sync with those in `setup.cfg`, otherwise it will break. Activating an automated dependency generator which is available only on Fedora does not relieve me of this, since I still have to do it for openSUSE.

Being explicit about the dependencies adds information value to the spec file, too, because at once glance I can tell you which RPM package meets the dependency on Fedora.

Conan-Kudo:
> Fedora offers an automated dependency generator. openSUSE does not. It's confusing to list Python dependencies only for one, but not the other.

This is not true. [openSUSE has had it for a few months now](https://github.com/openSUSE/python-rpm-macros/commit/8e390130273e5c31158c8c829970d378d62e74ae).

The macros are in place for Leap 15.2 as well, and I've [filed a bug to get the Provides generation enabled](https://bugzilla.opensuse.org/show_bug.cgi?id=1164533) for them to be effective there too.

Conan-Kudo:
> Being explicit about the dependencies adds information value to the spec file, too, because at once glance I can tell you which RPM package meets the dependency on Fedora, whereas setup.cfg does not tell me this information.

I'm not sure this is a good reason, because philosophically, RPM packages rely heavily on generators to ensure that the _correct_ dependencies are present in a form that is commonly consumable. This adds a mechanism in which you can rely on common names across distributions for Python modules, in the same way we do for libraries, Perl modules, Ruby gems, etc.

smithfarm:
@Conan-Kudo Cool! I wasn't aware of this. Once the supported openSUSE version (i.e. Leap 15.1 ATM) has a recent-enough rpm, remind me and I'll enable the automated dependency generation.

PR #144, opened 2020-03-06
spec: always install sesdev-qa RPM along with sesdev

smithfarm:
We want "sesdev create --qa-test" and "sesdev qa-test" to work always.
Right now, it's too easy to install just sesdev without sesdev-qa.

Signed-off-by: Nathan Cutler <ncutler@suse.com>

PR #145, opened 2020-03-06
Revert: "seslib: add SES7 Internal Media when --qa-test given"

smithfarm:
We no longer try to install the sesdev-qa package on the VMs. Instead, it gets installed on the host where sesdev is run, and the QA script files get copied to the VMs during the provisioning phase.


smithfarm:
merged into https://github.com/SUSE/sesdev/pull/144

PR #151, opened 2020-03-10
tests: enable unit testing via tox

smithfarm:

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ the VMs and run the deployment scripts.
       * [Install KVM/QEMU and Libvirt](#install-kvmqemu-and-libvirt-1)
       * [Install sesdev from package](#install-sesdev-from-package-1)
    * [Install sesdev from source](#install-sesdev-from-source)
-      * [Linting](#linting)
+      * [Running the unit tests](#running-the-unit-tests)
 * [Usage](#usage)
    * [Create/Deploy cluster](#createdeploy-cluster)
       * [Custom zypper repos](#custom-zypper-repos)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -164,20 +164,53 @@ Remember to re-run `pip install --editable .` after each git pull.
...
+
+```
+zypper -n install python3-tox
+```
+
+Then, execute the following commands in the top-level of your local git clone
+to install the dependencies, including test dependencies:
+
+```
+pip install --editable ./[dev]
```
toabctl: is `/` needed here?

README.md
```diff
@@ -164,20 +164,53 @@ Remember to re-run `pip install --editable .` after each git pull.
...
 ```
-zypper -n install python3-tox python3-pylint
+
+Finally, inspect the list of testing environments in `tox.ini` and choose one or
+more that you are interested in. Here is an example, but the actual output might
+be different:
+
 ```
+$ grep envlist tox.init
```
toabctl: This is `tox --listenv`

tests/test_pytest_smoke.py
```diff
@@ -0,0 +1,4 @@
+import pytest
```
toabctl: can you remove `_pytest` from the filename? not needed.

tests/test_pytest_smoke.py
```diff
@@ -0,0 +1,4 @@
...
+
+def test_pytest_smoke():
```
toabctl: can you remove `_pytest` from the function name? not needed

setup.cfg
```diff
@@ -48,8 +48,8 @@ console_scripts =
 [options.extras_require]
 dev =
     pytest
-    pycodestyle==2.5.0
-    pylint==2.4.4
+    pycodestyle
```
toabctl: I like the commit message! :)

README.md
smithfarm: Don't know. I tested this form of the command and it works.

tests/test_pytest_smoke.py
smithfarm: It's needed for me :-) This file contains a "pytest smoke test" - i.e. if the test passes, we know pytest is working.

tests/test_pytest_smoke.py
smithfarm: negative. see the other comment

README.md
smithfarm: changing to `tox --listenvs`

tests/test_pytest_smoke.py
toabctl: yeah. but usually python test files are just called `test_....py`. Independent from the test framework and test runner we use.

smithfarm: In this project, I prefer to be explicit that we are using `pytest`

smithfarm: (note, I'm not suggesting that we name all the files `test_pytest_....py` :-) Just this one.

smithfarm:
@toabctl Would `test_testing_subsystem_smoke` be less unacceptable for you?

tests/test_pytest_smoke.py
toabctl: ok :)

Issue #152, opened 2020-03-10
Jenkins runs sesdev with "--image-path null"


smithfarm:
The new automated deployment test has a glitch:

```
+ sesdev create octopus --non-interactive --image-path null --qa-test --single-node mini
=== Creating deployment with the following configuration ===
1 Deployment VMs:
  -- admin:
     - OS:               leap-15.2
     - ses_version:      octopus
     - deployment_tool:  orchestrator
     - roles:            ['mds', 'igw', 'storage', 'prometheus', 'grafana', 'ganesha', 'mon', 'rgw', 'mgr', 'admin']
     - fqdn:             admin.mini.com
     - public_address:   10.20.47.200
     - cpus:             2
     - ram:              4G
     - storage_disks:    4
                         (device names will be assigned by vagrant-libvirt)
     - repo_priority:    True
     - qa_test:          True
     - image_path:       null
```

toabctl:
I'll take a look into that.

smithfarm:
@toabctl Actually, the `--image-path` is unnecessary, since sesdev already defaults to the path Jenkins is supposed to be specifying. See #153.

toabctl:
@smithfarm if we want to use the job for other branches, we need to have this configurable I think. Or just rely on the defaults from sesdev?

smithfarm:
> if we want to use the job for other branches, we need to have this configurable I think

Which branches are you referring to? 

> Or just rely on the defaults from sesdev?

With the Jenkins job as it is now, it should use the default image path. Maybe we could drop this parameter now, and make it configurable later when there is a use case for it?

It **does** make sense for it to specify `--ceph-salt-repo https://github.com/ceph/ceph-salt --ceph-salt-branch master`, though.

PR #153, opened 2020-03-10
tests: tell Jenkins not to pass --image-path to "sesdev create octopus"

smithfarm:
The idea was for Jenkins to run "sesdev create octopus" with the
following command-line option:

--image-path registry.opensuse.org/filesystems/ceph/master/upstream/images/ceph/ceph

but:

1. it wasn't working
2. it's not needed because
   "registry.opensuse.org/filesystems/ceph/master/upstream/images/ceph/ceph"
   is already the default image path for "sesdev create octopus".

Fixes: https://github.com/SUSE/sesdev/issues/152

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -48,7 +48,7 @@ def zypper_add_repos(repo_list) {
     }
 }
 
-def sesdev_create_cmd = "sesdev create octopus --non-interactive ${sesdev_create_ceph_salt_params()} --image-path ${params.SESDEV_CEPH_CONTAINER_IMAGE} --qa-test --single-node mini"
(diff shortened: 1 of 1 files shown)


toabctl:
I would prefer to keep the image configurable. I can take a look

smithfarm:
@toabctl If you set an image path different from the default, the `--qa-test` will not pass. You would need to specify a `--repo` as well.

smithfarm:
Anyway, I'll be happy for you to take over :-)

smithfarm:
@toabctl I'm mystified why this change causes Jenkins to emit the command:

```
+ sesdev create octopus --non-interactive --qa-test --single-node mini
```

toabctl:
hm. Maybe we just add a free form `extra_params` parameter which is empty by default?

smithfarm:
closing in favor of #154

PR #154, opened 2020-03-10
Jenkinsfile: Use configured image

toabctl:
There is a Jenkins parameter for using a specific image. That
parameter is currently not used.

Fixes: #152

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -34,21 +34,12 @@ def cloud_params = [
     ]
 ]
 
-def sesdev_create_ceph_salt_params() {
-    if (params.CEPH_SALT_FROM_GIT_MASTER == true) {
-        return "--ceph-salt-repo https://github.com/ceph/ceph-salt.git --ceph-salt-branch master"
-    } else {
-        return ""
-    }
-}
-
 def zypper_add_repos(repo_list) {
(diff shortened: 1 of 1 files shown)


smithfarm:
Just to echo what I've written elsewhere...

There are currently two "versions" that use `--image-path`: octopus and ses7.

In order for the test to be valid, the version of cephadm and the version of ceph inside the container should be the same, and we have a QA test to enforce this.

Each "version" (octopus and ses7) has:

* a default set of repos for installing cephadm
* a default container image path

The defaults for these two are chosen to maximize the chances that

* a relevant (recent) version of cephadm+container image will be tested
* the version numbers will match

For these reasons, I don't think it makes sense for the Jenkins test to specify `--image-path` without also specifying `--repo` (these two go together).

smithfarm:
`The CEPH_SALT_FROM_GIT_MASTER is the default and it currently doesn't
make sense to have that configurable. So drop it.`

This is not true. AFAIK we need to always pass `--ceph-salt-repo https://github.com/ceph/ceph-salt.git --ceph-salt-branch master` explicitly. Otherwise we will get whichever version of ceph-salt that happens to be in the repos.

toabctl:
@smithfarm updated the commit msg.

smithfarm:
Thanks, @toabctl

Issue #157, opened 2020-03-12
Bootstrap minion doesn't have admin role


bk201:
After https://github.com/ceph/ceph-salt/pull/121, deploying octopus fails with:

```
    admin:   o- Containers .............................................................................................................. [...]
    admin:   | o- Images ................................................................................................................ [...]
    admin:   |   o- ceph ............................................................... [172.17.0.1:5001/ceph/daemon-base:latest-master-devel]
    admin:   o- Deployment .............................................................................................................. [...]
    admin:   | o- Bootstrap ......................................................................................................... [enabled]
    admin:   | o- Dashboard ............................................................................................................. [...]
    admin:   | | o- password ............................................................................................................ [***]
    admin:   | | o- username .......................................................................................................... [admin]
    admin:   | o- Mgr ............................................................................................................... [enabled]
    admin:   | o- Mon ............................................................................................................... [enabled]
    admin:   | o- OSD ............................................................................................................... [enabled]
    admin:   o- SSH ............................................................................................................ [Key Pair set]
    admin:   | o- Private_Key ............................................................... [cd:5f:30:e0:12:67:46:f9:97:bd:94:49:1e:66:0c:04]
    admin:   | o- Public_Key ................................................................ [cd:5f:30:e0:12:67:46:f9:97:bd:94:49:1e:66:0c:04]
    admin:   o- Storage ................................................................................................................. [...]
    admin:   | o- Drive_Groups ............................................................................................................ [1]
    admin:   |   o- {"testing_dg_admin": {"host_pattern": "admin*", "data_devices": {"all": true}}} ..................................... [...]
    admin:   o- System_Update ........................................................................................................... [...]
    admin:   | o- Packages ......................................................................................................... [disabled]
    admin:   | o- Reboot ........................................................................................................... [disabled]
    admin:   o- Time_Server ......................................................................................................... [enabled]
    admin:     o- External_Servers ........................................................................................................ [1]
    admin:     | o- 0.pt.pool.ntp.org ................................................................................................... [...]
    admin:     o- Server_Hostname ............................................................................................ [admin.mini.com]
    admin: ++ ceph-salt status
    admin: Bootstrap minion must be 'Admin'```

PR #158, opened 2020-03-12
seslib: Set admin roles for octopus

bk201:
Depends on: https://github.com/ceph/ceph-salt/pull/121

An example is also added to README.

Fixes: #157
Signed-off-by: Kiefer Chang <kiefer.chang@suse.com>

### DIFF
--- a/seslib/templates/ceph-salt/ceph_salt_deployment.sh.j2
+++ b/seslib/templates/ceph-salt/ceph_salt_deployment.sh.j2
@@ -51,6 +51,9 @@ exit 0
 
 {% for node in nodes %}
 ceph-salt config /Cluster/Minions add {{ node.fqdn }}
+{% if node.has_role('admin') %}
+ceph-salt config /Cluster/Roles/Admin add {{ node.fqdn }}
+{% endif %}
 {% if node.has_role('mon') %}
 ceph-salt config /Cluster/Roles/Mon add {{ node.fqdn }}
 {% endif %}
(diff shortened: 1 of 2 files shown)


seslib/templates/ceph-salt/ceph_salt_deployment.sh.j2
```diff
@@ -51,6 +51,9 @@ exit 0
 
 {% for node in nodes %}
 ceph-salt config /Cluster/Minions add {{ node.fqdn }}
+{% if node.has_role('admin') %}
```
ricardoasmarques: This means that the default `sesdev create octopus <name>` will not work because we need one minion to be Admin, Mon, and Mgr at the same time.

I see some possible solutions:

1) Make all minions `Admin` (regardless of the roles specified in the `--roles')
2) Or have different default roles for octopus
3) Or what we have done in this PR

IMO, we can merge this PR and improve later, but I would like to hear which option do you guys prefer.

smithfarm: @ricardoasmarques I agree we can merge this PR as it is, after the corresponding ceph-salt PR is merged.

Getting octopus to work with the default roles will require some extra work. I'll open an issue for it.

ricardoasmarques:
FYI https://github.com/ceph/ceph-salt/pull/121 has been merged

smithfarm:
@bk201 Please rebase - I just made some changes to README.md that caused a conflict here. Sorry about that!

smithfarm:
Thanks, @bk201 !

bk201:
> Nevermind - this PR was urgently needed, so I did the rebase.

Thanks for that!

Issue #160, opened 2020-03-12
Change "admin role" semantics


smithfarm:
Until https://github.com/ceph/ceph-salt/pull/121 it made sense to call the Salt Master node the "admin" node and enforce that there is only one "admin" node in the cluster.

After https://github.com/ceph/ceph-salt/pull/121, ceph-salt is using more or less the same semantics as DeepSea's `policy.cfg` - i.e. the "master" role means that this node is the Salt Master and the "admin" role means that this node should get `ceph.conf` and the admin keyring.

So, for all deployment versions except `caasp4` it now makes sense to:

* change current "admin" role to "master" everywhere
* newly give all nodes the "admin" role, regardless of whether or not the user specified it

PR #161, opened 2020-03-12
Change "admin" role semantics and make roles configurable

smithfarm:
Fixes: https://github.com/SUSE/sesdev/issues/160

---

To Do:

- [x] test does not break `sesdev list`
- [x] test does not break `sesdev destroy`
- [x] test does not break `sesdev create octopus`
- [x] test with `version_default_roles` stanza in `config.yaml`
- [x] test does not break `sesdev ssh` on newly deployed cluster
- [x] test does not break `sesdev create nautilus`
- [x] test does not break `sesdev ssh DEP_ID admin` on legacy cluster
- [x] test does not break `sesdev create caasp4`


ricardoasmarques:
I found the following issue while testing:
```
# sesdev list        
Setting 'qa_test' is not known - please open a bug report!
```

smithfarm:
@ricardoasmarques Good catch. Added a commit to address this problem.

PR #165, opened 2020-03-13
Add a bootstrap.sh script and document how to use it

smithfarm:

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+if [ -d ./sesdev ] ; then
+    true
+else
+    >&2 echo "The working directory, $(pwd), does not look like a sesdev git clone."
+    >&2 echo "Bailing out!"
+    exit 1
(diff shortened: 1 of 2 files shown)


bootstrap.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
toabctl: Why do we need that script? It just duplicates `tox -epy37; source .tox/py37/bin/activate`

smithfarm: It's pretty standard to have a bash script for installing dependencies. Ceph has one, teuthology has one, plenty of projects have one.

Not everyone is a Python whiz. Asking users to run "./bootstrap.sh" is potentially much less intimidating than asking them to run "tox -epy37; source .tox/py37/bin/activate".

Having a bootstrap script also makes the README much simpler.

Beyond that, having the commands in a script, as opposed to in a README file, makes the installation procedure much easier to test.

PR #166, opened 2020-03-13
provision.sh: add SUSE:CA repo on ses5

smithfarm:
After some recent refactoring, the `ca-certificates-suse` package was no longer getting installed on `ses5` and `sesdev` was not raising it as an error.

This patchset fixes both issues.


smithfarm:
I have tested this. It works great.

Somebody, please review.

Issue #155, opened 2020-03-11
bug: provision destination needs trailing slash if directories are copied


jschmid1:
when invoking sesdev with:

`sesdev create ses7 --single-node --qa-test test_ses7 `

it will fail on the provisioning (file) phase with:

```
DEBUG ssh: Uploading: $sesdev_path_to_qa to /home/vagrant/sesdev-qa
DEBUG ssh: Re-using SSH connection.
ERROR warden: Error occurred: scp: error: unexpected filename: .
```

This is probably due to how the `destination` key in the Vagrantfile is implemented.

`node.vm.provision "file", source: "{{ sesdev_path_to_qa }}", destination: "/home/vagrant/sesdev-qa"` <- no trailing slash will look for a `.` in `sesdev_path_to_qa`

Adding the trailing slash actually fixes this issue by copying the entire directory `sesdev_path_to_qa` under `/home/vagrant/sesdev-qa/`

-> `/home/vagrant/sesdev-qa/qa/$content_of_qa_dir`

smithfarm:
@jschmid1 Are you running sesdev from source or from RPM?

smithfarm:
Also, which version of sesdev - include output of `sesdev --version`

smithfarm:
Here's what I see on the latest version of sesdev (installed from source):

```
==> admin: Running provisioner: file...
    admin: /home/smithfarm/sesdev/sesdev/../qa/ => /home/vagrant/sesdev-qa
==> admin: Running provisioner: shell...
    admin: Running: /tmp/vagrant-shell20200311-17616-to8qgo.sh
    admin: ++ ls -lR /home/vagrant
    admin: /home/vagrant:
    admin: total 8
    admin: drwxr-xr-x 2 vagrant vagrant 4096 Oct  9 04:36 bin
    admin: drwxr-xr-x 3 vagrant vagrant 4096 Mar 11 15:19 sesdev-qa
    admin:
    admin: /home/vagrant/bin:
    admin: total 0
    admin:
    admin: /home/vagrant/sesdev-qa:
    admin: total 12
    admin: drwxr-xr-x 2 vagrant vagrant 4096 Mar 11 15:19 common
    admin: -rwxr-xr-x 1 vagrant vagrant 3743 Mar 11 15:19 health-ok.sh
    admin: -rw-r--r-- 1 vagrant vagrant 1415 Mar 11 15:19 README.rst
    admin:
    admin: /home/vagrant/sesdev-qa/common:
[...]

PR #167, opened 2020-03-13
Vagrantfile,sesdev.spec: require vagrant > 2.2.2

smithfarm:
sesdev is known to require a recent vagrant version, but it's not
entirely clear how recent.

From https://github.com/SUSE/sesdev/issues/155 we know that 2.2.0 will
not work.

Virtualization:vagrant/vagrant was updated to 2.2.2 on January 4, 2019.
This is probably also too old, so enforce vagrant > 2.2.2

### DIFF
--- a/sesdev.spec
+++ b/sesdev.spec
@@ -46,7 +46,7 @@ Requires:       python3-pyyaml >= 3.13
 Requires:       python3-click >= 6.7
 Requires:       python3-pycryptodomex >= 3.4.6
 Requires:       python3-setuptools
-Requires:       vagrant
+Requires:       vagrant > 2.2.2
 Requires:       vagrant-libvirt
 Requires:       sesdev-qa
 
--- a/seslib/templates/Vagrantfile.j2
+++ b/seslib/templates/Vagrantfile.j2
@@ -1,6 +1,8 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+Vagrant.require_version "> 2.2.2"
+
 Vagrant.configure("2") do |config|
   config.ssh.insert_key = false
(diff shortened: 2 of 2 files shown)


bootstrap.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
```
toabctl: Can we just use https://www.vagrantup.com/docs/vagrantfile/vagrant_version.html instead of this huge shell script?

smithfarm:
```
Do you want to continue with the deployment? [Y/n]: 
=== Running shell command ===
vagrant up
Vagrant failed to initialize at a very early stage:

This Vagrant environment has specified that it requires the Vagrant
version to satisfy the following version requirements:

  >= 2.3

You are running Vagrant 2.2.7, which does not satisfy
these requirements. Please change your Vagrant version or update
the Vagrantfile to allow this Vagrant version. However, be warned
that if the Vagrantfile has specified another version, it probably has
good reason to do so, and changing that may cause the environment to
not function properly.
```

@toabctl Nice! Thanks for the tip. That will make the shell script much smaller ;-)

toabctl:
The build seems to hang at `cephadm bootstrap`. I saw that yesterday, too. See https://ceph-ci.suse.de/blue/organizations/jenkins/sesdev-integration/detail/PR-167/5/pipeline#step-88-log-1891

smithfarm:
> The build seems to hang at `cephadm bootstrap`. I saw that yesterday, too. See https://ceph-ci.suse.de/blue/organizations/jenkins/sesdev-integration/detail/PR-167/5/pipeline#step-88-log-1891

Unrelated

Issue #172, opened 2020-03-19
ses5 deployment tries to remove ntp and fails when it isn't there


smithfarm:
Apparently, the SLE-12-SP3 Vagrant box used by sesdev used to have the `ntp` package installed, but that is no longer the case. As a result, the `ses5` deployment fails with:

```
    master: ++ zypper -n rm ntp
    master: Loading repository data...
    master: Warning: No repositories defined. Operating only with the installed resolvables. Nothing can be installed.
    master: Reading installed packages...
    master: 'ntp' not found in package names. Trying capabilities.
    master: No provider of 'ntp' found.
    master: Resolving package dependencies...
    master: Nothing to do.
Command '['vagrant', 'up']' failed: ret=1 stderr:
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

This is a consequence of #166 which added `set -e` at the top of the provisioning script.

PR #173, opened 2020-03-19
provision.sh: do not fail ses5 deployment if ntp not installed

smithfarm:
Fixes: https://github.com/SUSE/sesdev/issues/172

### DIFF
--- a/seslib/templates/provision.sh.j2
+++ b/seslib/templates/provision.sh.j2
@@ -108,7 +108,7 @@ systemctl start salt-minion
 {% endif %}
 
 {% if version == 'ses5' %}
-zypper -n rm ntp
+zypper -n rm ntp || true
 {% endif %}
 
 touch /tmp/ready


smithfarm:
I have tested this. It works great.

Somebody, please review.

seslib/templates/provision.sh.j2
```diff
@@ -108,7 +108,7 @@ systemctl start salt-minion
 {% endif %}
 
 {% if version == 'ses5' %}
-zypper -n rm ntp
+zypper -n rm ntp || true
```
toabctl: Why not removing the whole block if the package is not installed?

smithfarm: because it might get installed in the future?

toabctl: I thought the package got recently removed from the vagrant box. So I would not assume that it will be added again...

smithfarm: I think it's better not to assume anything. We know we don't want the package to be installed, so this line ensures (100%) that it will not be installed.

Issue #75, opened 2020-02-03
feature request: "sesdev report DEPLOYMENT_ID NODE_ID" to automate supportconfig on SLE


smithfarm:
In cases when deployment fails and a supportconfig is needed, there could be a

    sesdev report DEPLOYMENT_ID NODE_ID

feature which would take care of installing `supportconfig` on the cluster and running it to get data for debugging.

smithfarm:
Note that supportconfig is only available on SLE (openSUSE doesn't have it).

PR #176, opened 2020-03-19
Implement feature: "sesdev supportconfig DEPLOYMENT_ID NODE_ID"

smithfarm:
Fixes: https://github.com/SUSE/sesdev/issues/75

---

To Do:

- [ ] assert that it doesn't break `sesdev ssh --recursive`

### DIFF
--- a/seslib/__init__.py
+++ b/seslib/__init__.py
@@ -19,7 +19,7 @@ from .exceptions import DeploymentDoesNotExists, VersionOSNotSupported, SettingT
                         ServicePortForwardingNotSupported, DeploymentAlreadyExists, \
                         ServiceNotFound, ExclusiveRoles, RoleNotSupported, CmdException, \
                         VagrantSshConfigNoHostName, ScpInvalidSourceOrDestination, \
-                        TooManyMasters, SettingNotKnown
+                        TooManyMasters, SettingNotKnown, SupportconfigOnlyOnSLE
 
 
 JINJA_ENV = Environment(loader=PackageLoader('seslib', 'templates'), trim_blocks=True)
@@ -1361,7 +1361,7 @@ class Deployment():
     def ssh(self, name, command):
         tools.run_interactive(self._ssh_cmd(name, command))
 
-    def _scp_cmd(self, recursive, source, destination):
+    def _scp_cmd(self, source, destination, recurse=False):
         host_is_source = False
         host_is_destination = False
         name = None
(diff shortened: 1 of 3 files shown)


smithfarm:
@tserong @toabctl @ricardoasmarques @capri1989 This runs `supportconfig` on a node deployed by sesdev, downloads the resulting tarball to the host, and deletes the tarball on the node (so it doesn't get downloaded again if the command is run multiple times).

Would be very happy if you gave it a spin and a review!

It occurred to me that `sesdev supportconfig` might be a better name than `sesdev report`. What do you think?

ricardoasmarques:
> @tserong @toabctl @ricardoasmarques @capri1989 This runs `supportconfig` on a node deployed by sesdev, downloads the resulting tarball to the host, and deletes the tarball on the node (so it doesn't get downloaded again if the command is run multiple times).
> 
> Would be very happy if you gave it a spin and a review!
> 
> It occurred to me that `sesdev supportconfig` might be a better name than `sesdev report`. What do you think?

`sesdev supportconfig` is indeed more clear/explicit IMO.

seslib/__init__.py
```diff
@@ -1409,8 +1409,23 @@ def _scp_cmd(self, recursive, source, destination):
 
         return _cmd
 
-    def scp(self, recursive, source, destination):
-        tools.run_interactive(self._scp_cmd(recursive, source, destination))
+    def scp(self, source, destination, recurse=False):
+        tools.run_interactive(self._scp_cmd(source, destination, recurse=recurse))
+
+    def report(self, log_handler, name):
+        if self.settings.os.startswith("sle"):
```
ricardoasmarques: nit: Perhaps:
```
if not self.settings.os.startswith("sle"):
    raise SupportconfigOnlyOnSLE()
```
(wihout the `else`) is easier to read

smithfarm: Replaced `pass` with a log message to legitimize this construct. (I find positive conditional expressions easier/faster to grasp.)

seslib/exceptions.py
```diff
@@ -113,3 +113,11 @@ class SettingNotKnown(SesDevException):
     def __init__(self, setting):
         super(SettingNotKnown, self).__init__(
             "Setting '{}' is not known - please open a bug report!".format(setting))
+
+
+class SupportconfigOnlyOnSLE(SesDevException):
+    def __init__(self):
+        super(SupportconfigOnlyOnSLE, self).__init__(
+            "sesdev report depends on supportconfig, which is available only on "
```
ricardoasmarques: It's no longer `sesdev report`

smithfarm: fixed

PR #177, opened 2020-03-20
provision.sh: Work around podman/runc issue

smithfarm:
Works-around: https://github.com/ceph/ceph-salt/issues/130

### DIFF
--- a/seslib/templates/provision.sh.j2
+++ b/seslib/templates/provision.sh.j2
@@ -100,7 +100,14 @@ hostnamectl set-hostname {{ node.name }}
 
 {% if not suma %}
 zypper -n install salt-minion
-sed -i 's/^#master:.*/master: {{ master.name }}/g' /etc/salt/minion
+sed -i -e 's/^#master:.*/master: {{ master.name }}/g' /etc/salt/minion
+{% if version == 'octopus' or version == 'ses7' %}
+#
+# FIXME: work around bsc#1167218 - remove this if block once that bug is resolved
+#
+sed -i -e 's/^Type=notify/#Type=notify/' /usr/lib/systemd/system/salt-minion.service
+sed -i -e 's/^NotifyAccess=all/#NotifyAccess=all/' /usr/lib/systemd/system/salt-minion.service
+{% endif %}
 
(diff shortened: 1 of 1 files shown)


sebastian-philipp:
What about adding this workaround to ceph-salt? ceph/ceph-salt#130 is not limited to sesdev!

smithfarm:
> What about adding this workaround to ceph-salt? [ceph/ceph-salt#130](https://github.com/ceph/ceph-salt/issues/130) is not limited to sesdev!

That's not a sesdev issue at all, actually. But it affects sesdev and the workaround costs nothing to carry here.

smithfarm:
I was confident this workaround would work, and it does work, so I'm merging it now.

bk201:
Tested and worked.

seslib/templates/provision.sh.j2
```diff
@@ -100,7 +100,14 @@ hostnamectl set-hostname {{ node.name }}
 
 {% if not suma %}
 zypper -n install salt-minion
-sed -i 's/^#master:.*/master: {{ master.name }}/g' /etc/salt/minion
+sed -i -e 's/^#master:.*/master: {{ master.name }}/g' /etc/salt/minion
+{% if version == 'octopus' or version == 'ses7' %}
+#
+# FIXME: work around bsc#1167218 - remove this if block once that bug is resolved
+#
+sed -i -e 's/^Type=notify/#Type=notify/' /usr/lib/systemd/system/salt-minion.service
```
toabctl: Why does the commit message mention podman/runc if this is changing salt-minion service? The bug is also not very helpful.

tserong:
I know I'm late to the party on this (sorry!) but I'd really rather this was reverted and we just run with the workaround in https://github.com/ceph/ceph-salt/pull/134 (rationale: https://bugzilla.suse.com/show_bug.cgi?id=1149954#c4 says "We use "Type=notify", because we indeed notify[1] systemd about the readiness of the process.")

smithfarm:
> I'd really rather this was reverted 

Done!

tserong:
Thanks! :-)

PR #179, opened 2020-03-24
Jenkinsfile.integration: Parametrize ceph-salt repo and branch

toabctl:
That way, we can trigger the job also manually or from other jobs with
a specific ceph-salt repo.


smithfarm:
```
+ sesdev create octopus --non-interactive --ceph-salt-repo https://github.com/ceph/ceph-salt.git --ceph-salt-branch master --qa-test --single-node mini
```

It doesn't break the current use case, so LGTM.

PR #182, opened 2020-03-24
Jenkins improvements

toabctl:


smithfarm:
@toabctl Can you figure out why, in this PR, the ceph-salt installation from source causes salt 2019 to be installed from source into the virtualenv? Compare with other runs, where this does not happen.

toabctl:
@smithfarm I triggered a job manually with another salt repo. Let me retrigger without any parameter changes...

Jenkinsfile.integration
```diff
@@ -40,14 +40,19 @@ def zypper_add_repos(repo_list) {
...
     cmd = "source venv/bin/activate; sesdev create octopus --non-interactive"
     if (!ceph_salt_git_repo.isEmpty()) {
         if (ceph_salt_git_branch.isEmpty()) {
             error('ceph-salt git repo set, but no ceph-salt git branch given')
         }
         cmd += " --ceph-salt-repo ${ceph_salt_git_repo} --ceph-salt-branch ${ceph_salt_git_branch}"
     }
+
+    if (!extra_repo_url.isEmpty()) {
+        cmd += " --repo ${extra_repo_url}"
```
smithfarm: probably you want to add `--no-repo-priority` here. Without that, the new repo gets added with elevated zypper priority, which can cause surprising and unexpected behavior, depending on what the repo is.

Jenkinsfile.integration
```diff
@@ -65,6 +70,8 @@ pipeline {
                description: 'ceph-salt git repository to use for installation. If empty, the RPM package is used')
         string(name: 'CEPH_SALT_GIT_BRANCH', defaultValue: 'master',
                description: 'ceph-salt git branch to use for installation. This parameter is only used when CEPH_SALT_GIT_REPO is set')
+        string(name: 'EXTRA_REPO_URL', defaultValue: '',
+               description: 'Extra zypper repo url that will be added to all nodes')
```
smithfarm: what if I want to add two extra repos? or three? :-)

toabctl: It's not obvious if I can add the flag `--repo` multiple times from `sesdev -h`. Can I?

smithfarm: yes, you can

Jenkinsfile.integration
toabctl: I would like to keep the default. And having an elevated zypper prio is actually what I would expect.

Jenkinsfile.integration
toabctl: Fixed. parameter accepts now a space separated list of urls.

smithfarm:
@toabctl The new "pr-merge" Jenkins job is a great idea, but it's producing suspicious results? The "Check out from version control" step has the following messages at the end of the output:

```
First time build. Skipping changelog.
fatal: bad object 9549fdaf59dd7308c16e8c897613ae179727841b
```

Do you know what they mean?

toabctl:
> @toabctl The new "pr-merge" Jenkins job is a great idea, but it's producing suspicious results? The "Check out from version control" step has the following messages at the end of the output:

Oh. Ignore that job. I'm experimenting with ceph-salt testing. Sorry for the noise.

toabctl:
> nit: Maybe rename `extra_repo_url` parameter to `extra_repo_urls`

Done

PR #183, opened 2020-03-24
Set "osd crush chooseleaf type = 0" via bootstrap ceph.conf in very small clusters

ricardoasmarques:
Set "osd crush chooseleaf type = 0"

After https://github.com/ceph/ceph-salt/pull/129

### DIFF
--- a/seslib/templates/ceph-salt/ceph_salt_deployment.sh.j2
+++ b/seslib/templates/ceph-salt/ceph_salt_deployment.sh.j2
@@ -88,6 +88,10 @@ ceph-salt config /Storage/Drive_Groups add value="{\"service_type\": \"osd\", \"
 
 {% endif %}
 {% endfor %}
+{% if storage_nodes < 3 %}
+ceph-salt config /Deployment/Bootstrap_Ceph_Conf add global
+ceph-salt config /Deployment/Bootstrap_Ceph_Conf/global set osd crush chooseleaf type = 0
+{% endif %}
 {% endif %} {# if ceph_salt_deploy_osds #}
 
 ceph-salt config /Deployment/Dashboard/username set admin


seslib/templates/ceph-salt/ceph_salt_deployment.sh.j2
```diff
@@ -82,12 +82,18 @@ ceph-salt config /Deployment/Mgr enable
...
 # OSDs drive groups spec for each node
+{% set storage_nodes = [] %}
 {% for node in nodes %}
 {% if node.has_role('storage') %}
+{% if storage_nodes.append(node) %}{% endif %}
 ceph-salt config /Storage/Drive_Groups add value="{\"service_type\": \"osd\", \"placement\": {\"host_pattern\": \"{{ node.name }}*\"}, \"service_id\": \"testing_dg_{{ node.name }}\", \"data_devices\": {\"all\": True}}"
 
 {% endif %}
 {% endfor %}
+{% if storage_nodes | length == 1 %}
```
smithfarm: I think failure domain should be "osd" even if there are two storage nodes, so maybe

```
{% if storage_nodes | length < 3 %}
```

seslib/templates/ceph-salt/ceph_salt_deployment.sh.j2
```diff
@@ -82,12 +82,18 @@ ceph-salt config /Deployment/Mgr enable
 ceph-salt config /Deployment/OSD enable
 
 # OSDs drive groups spec for each node
+{% set storage_nodes = [] %}
```
smithfarm: this line can be omitted because `storage_nodes` is already set to the number of storage nodes:

https://github.com/SUSE/sesdev/blob/master/seslib/__init__.py#L1081

seslib/templates/ceph-salt/ceph_salt_deployment.sh.j2
```diff
@@ -82,12 +82,18 @@ ceph-salt config /Deployment/Mgr enable
...
 {% for node in nodes %}
 {% if node.has_role('storage') %}
+{% if storage_nodes.append(node) %}{% endif %}
```
smithfarm: this line can be omitted because `storage_nodes` is already set to the number of storage nodes:

https://github.com/SUSE/sesdev/blob/master/seslib/__init__.py#L1081

ricardoasmarques:
@smithfarm nice catch, it's fixed now.

toabctl:
Having a short description *why* this is needed in the commit message would be nice.

smithfarm:
> Having a short description _why_ this is needed in the commit message would be nice.

It's needed because, without it, `--single-node` will start to fail after https://github.com/ceph/ceph-salt/pull/129 is merged.

PR #191, opened 2020-03-25
ceph-salt: guard fetching of all GitHub PRs with CLI option

smithfarm:
Fetching of all GitHub PRs is needed by the ceph-salt Jenkins CI, but not (or, at least, not in the typical case) by ordinary users, so guard it with a command-line option that defaults to False.

### DIFF
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -67,6 +67,8 @@ def ceph_salt_options(func):
                      help='ceph-salt Git repo URL'),
         click.option('--ceph-salt-branch', type=str, default=None,
                      help='ceph-salt Git branch'),
+        click.option('--fetch-github-prs/--no-fetch-github-prs', default=False,
+                     help="Fetch all GitHub PRs when installing ceph-salt from source")
         click.option('--image-path', type=str, default=None,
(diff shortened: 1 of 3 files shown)


sesdev/__init__.py
```diff
@@ -67,6 +67,8 @@ def ceph_salt_options(func):
                      help='ceph-salt Git repo URL'),
         click.option('--ceph-salt-branch', type=str, default=None,
                      help='ceph-salt Git branch'),
+        click.option('--fetch-github-prs/--no-fetch-github-prs', default=False,
```
toabctl: I would call it `--ceph-salt-fetch-github-prs`. I know it's longer but then it's in line with `ceph-salt-branch` and `ceph-salt-repo`.

sesdev/__init__.py
```diff
@@ -67,6 +67,8 @@ def ceph_salt_options(func):
...
+                     help="Fetch all GitHub PRs when installing ceph-salt from source")
```
toabctl: Maybe mention that this is only used when `ceph-salt-repo` is set?

smithfarm:
closing in favor of #197

Issue #181, opened 2020-03-24
ses6: "cluster does not have a DNS domain configured" error in Stage 2 orchestration step parsing phase


jan--f:
Using this create command:
`sesdev create ses6 --libvirt-networks routed_all --domain virt1.home.fajerski.name --roles "[master, mon, mgr], [storage, mon, mgr],[storage,mon.mgr],[storage, mds],[storage,mds]" --os sles-15-sp1 --libvirt-storage-pool sesdev-nvme --num-disks 2 --ram 4 ses`

I end up with a cluster where the minion ids are `master`, `node1` and so on. This interferes with deepsea deployments, eg:
```
 master: ***** RUNNING stage.2 *******
    master: ++ echo ''
    master: ++ echo '***** RUNNING stage.2 *******'
    master: ++ deepsea stage run --simple-output ceph.stage.2
    master: Starting orchestration: ceph.stage.2
    master: Parsing orchestration ceph.stage.2 steps... 
    master: fail
    master: 
    master: An error occurred when rendering one of the following states:
    master:     - ceph.ssl (/srv/salt/ceph/ssl)
    master: Error description:
    master:   - Rendering SLS 'base:ceph.ssl.default' failed: Jinja error: According to the 'domain' grain, the cluster does not have a DNS domain configured
    master:     Traceback (most recent call last):
    master:       File "/usr/lib/python3.6/site-packages/salt/utils/templates.py", line 392, in render_jinja_tmpl
    master:         output = template.render(**decoded_context)
    master:       File "/usr/lib/python3.6/site-packages/jinja2/asyncsupport.py", line 76, in render
    master:         return original_render(self, *args, **kwargs)
    master:       File "/usr/lib/python3.6/site-packages/jinja2/environment.py", line 1008, in render
    master:         return self.environment.handle_exception(exc_info, True)
    master:       File "/usr/lib/python3.6/site-packages/jinja2/environment.py", line 780, in handle_exception
    master:         reraise(exc_type, exc_value, tb)
    master:       File "/usr/lib/python3.6/site-packages/jinja2/_compat.py", line 37, in reraise
    master:         raise value.with_traceback(tb)
    master:       File "<template>", line 1, in top-level template code
    master:       File "/var/cache/salt/minion/extmods/modules/deepsea.py", line 97, in ssl_cert_cn_wildcard
    master:         raise Exception("According to the 'domain' grain, the cluster does"
[...]

PR #199, opened 2020-03-26
provisioning: set fqdn through vagrant and leave /etc/hosts alone

jan--f:
Signed-off-by: Jan Fajerski <jfajerski@suse.com>

Fixes: https://github.com/SUSE/sesdev/issues/181

Not sure why we only set the name as the hostname when we have the fqdn. Alos can't hurt to set the minion id explicitly.

### DIFF
--- a/seslib/templates/provision.sh.j2
+++ b/seslib/templates/provision.sh.j2
@@ -4,9 +4,6 @@ set -ex
 
 ls -lR /home/vagrant
 
-# remove the first line introduced by vagrant
-head -1 /etc/hosts | grep -q '127.*' && sed -i '1d' /etc/hosts
-
 {% for _node in nodes %}
 {% if _node.public_address %}
 echo "{{ _node.public_address }} {{ _node.fqdn }} {{ _node.name }}" >> /etc/hosts
--- a/seslib/templates/engine/libvirt/vagrant.node.j2
+++ b/seslib/templates/engine/libvirt/vagrant.node.j2
@@ -1,6 +1,6 @@
     {{ node.networks }}
 
-    node.vm.hostname = "{{ node.name }}"
+    node.vm.hostname = "{{ node.fqdn }}"
 
     node.vm.provider "libvirt" do |lv|
       lv.memory = {{ node.ram }}


jan--f:
hmm this ci failure is pretty unreadable...but going out on a limb and claim its not related...?

seslib/templates/provision.sh.j2
```diff
@@ -92,7 +92,7 @@ cat /home/vagrant/.ssh/id_rsa.pub >> /home/vagrant/.ssh/authorized_keys
 chmod 600 /home/vagrant/.ssh/id_rsa
 cp /home/vagrant/.ssh/id_rsa* /root/.ssh/
 cat /root/.ssh/id_rsa.pub >> /root/.ssh/authorized_keys
-hostnamectl set-hostname {{ node.name }}
+hostnamectl set-hostname {{ node.fqdn }}
```
sebastian-philipp: * `hostname` should return the bare hostname. 
* `hostname -f` should return the fqdn

Please correct me, if I'm wrong.

jan--f: >     * `hostname` should return the bare hostname.

I guess you mean the short hostname and that is returned by `hostname -f`

What we really need to do here is get salt to return the domain grain properly. But I guess we can explicitly set that instead or relying on salt.

toabctl:
Could you describe in the commit message why we want/need this please?

jan--f:
Ok this seems to do now what deepsea expects. What does everybody think?

```
master:~ # hostname
master.virt1.home.fajerski.name
master:~ # hostname -f
master.virt1.home.fajerski.name
master:~ # hostname -s
master
master:~ # salt '*' grains.get fqdn
master.virt1.home.fajerski.name:
    master.virt1.home.fajerski.name
node3.virt1.home.fajerski.name:
    node3.virt1.home.fajerski.name
node1.virt1.home.fajerski.name:
    node1.virt1.home.fajerski.name
node2.virt1.home.fajerski.name:
    node2.virt1.home.fajerski.name
node4.virt1.home.fajerski.name:
    node4.virt1.home.fajerski.name
master:~ # salt '*' grains.get domain
node3.virt1.home.fajerski.name:
    virt1.home.fajerski.name
node4.virt1.home.fajerski.name:
    virt1.home.fajerski.name
node1.virt1.home.fajerski.name:
    virt1.home.fajerski.name
node2.virt1.home.fajerski.name:
    virt1.home.fajerski.name
master.virt1.home.fajerski.name:
    virt1.home.fajerski.name
```

smithfarm:
The current code in master appears to work fine on openSUSE Leap 15.1 with Virtualization:vagrant, so would it be nice if we could pinpoint the environmental difference that is causing the issue on Tumbleweed?

Also, can anyone else reproduce the issue?

smithfarm:
Also, if there is any way to avoid it, I would rather not rely on steps that are not in the SES6 deployment documentation.

I'm not saying that there are no such steps currently in sesdev -- it's possible that there are -- but in general it seems that we should try to avoid adding more, if possible.

jan--f:
latest box shows the same issues for me. For the record, here's the error:
```
 master: An error occurred when rendering one of the following states:
    master:     - ceph.ssl (/srv/salt/ceph/ssl)
    master: 
    master: Error description:
    master:   - Rendering SLS 'base:ceph.ssl.default' failed: Jinja error: According to the 'domain' grain, the cluster does not have a DNS domain configured
    master:     Traceback (most recent call last):
    master:       File "/usr/lib/python3.6/site-packages/salt/utils/templates.py", line 392, in render_jinja_tmpl
    master:         output = template.render(**decoded_context)
    master:       File "/usr/lib/python3.6/site-packages/jinja2/asyncsupport.py", line 76, in render
    master:         return original_render(self, *args, **kwargs)
    master:       File "/usr/lib/python3.6/site-packages/jinja2/environment.py", line 1008, in render
    master:         return self.environment.handle_exception(exc_info, True)
    master:       File "/usr/lib/python3.6/site-packages/jinja2/environment.py", line 780, in handle_exception
    master:         reraise(exc_type, exc_value, tb)
    master:       File "/usr/lib/python3.6/site-packages/jinja2/_compat.py", line 37, in reraise
    master:         raise value.with_traceback(tb)
    master:       File "<template>", line 1, in top-level template code
    master:       File "/var/cache/salt/minion/extmods/modules/deepsea.py", line 97, in ssl_cert_cn_wildcard
    master:         raise Exception("According to the 'domain' grain, the cluster does"
    master:     Exception: According to the 'domain' grain, the cluster does not have a DNS domain configured
    master:     
    master:     ; line 1
    master:     
    master:     ---
    master:     {% set CN = salt['deepsea.ssl_cert_cn_wildcard']() %}    <======================
    master:     
    master:     create CA:
    master:       module.run:
    master:         - name: tls.create_ca
    master:         - ca_name: deepsea
    master:     [...]
    master:     ---
==> master: An error occurred. The error will be shown after all tasks complete.
An error occurred while executing multiple actions in parallel.
Any errors that occurred are shown below.

An error occurred while executing the action on the 'master'
machine. Please handle this error then try again:

The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```



> Also, if there is any way to avoid it, I would rather not rely on steps that are not in the SES6 deployment documentation.

Agreed. Though iiuc these are assumptions we make in DeepSea.

smithfarm:
Do we know how the `domain` grain gets populated?

jan--f:
So this seems to be related to the `--libvirt-network` option. salt does not deal well with this. `hostname` behaves as expected, but the minion_id and grains are not as expected. This is likely due to the default setup adding all hosts to `/etc/hosts`.

So I think this patch is a reasonable solution for this issue. Arguably this is closer to what would be done in the real world as opposed to populating `/etc/hosts`.

> Do we know how the `domain` grain gets populated?

We looked into this at some point. Iirc the result was that it checks in various places but does not query the network or ask `hostname -f`.

smithfarm:
@jan--f Would it make sense to guard this - with something like `{% if libvirt_network_option %}` - so the behavior only changes when `--libvirt-network` option is given on the command line?

smithfarm:
Also, does the same issue exist in `ses5` and `ses7` too, since these also use Salt?

seslib/templates/provision.sh.j2
smithfarm: Is this line required? I mean, will the issue be fixed even without this line?

sebastian-philipp:
ses7 is going to get interesting. Sage was against supporting "`hostname` returning the fqdn" a few months back, as that's what `-f` is for. see https://github.com/ceph/ceph/pull/33042  

If we want that, we have to pass `--allow-fqdn-hostnam` to `cephadm bootstrap` and make sure we fix all upcoming issues

jan--f:
> ses7 is going to get interesting. Sage was against supporting "`hostname` returning the fqdn" a few months back, as that's what `-f` is for. see [ceph/ceph#33042](https://github.com/ceph/ceph/pull/33042)
> 
> If we want that, we have to pass `--allow-fqdn-hostnam` to `cephadm bootstrap` and make sure we fix all upcoming issues, otherwise we're creating a health warning

hmm I don't fully understand this. cephadm does not use the hostname tool, but it uses `socket.gethostname`. Also I'm confused by the motivation for this PR...a) it doesn't warn but actually errors out, b) why can't cephadm just shorten the hostname if needed and use that as daemon ids?

jan--f:
> @jan--f Would it make sense to guard this - with something like `{% if libvirt_network_option %}` - so the behavior only changes when `--libvirt-network` option is given on the command line?
> 
> The rationale:
> 
> sesdev is a testing tool. Often when investigating a bug one does not know what the cause is and any part of the code _that is run_ is potentially the cause of the bug. By guarding this code, someone who does not use `--libvirt-network` who is investigating a bug would know that it isn't running in their case, so they wouldn't have to consider it as a potential cause of the bug.

I understand that sesdev is a testing tool first and foremost. My argument as to why this PR is reasonable is that we have certain apriori expectations to a salt cluster. That sesdev on ses5/ses6 fullfills these is merely a side effect of populating /etc/hosts in the default case, `--libvirt-networks` skips this step. This PR just adds other mechanisms to prepare the salt cluster as per our expectations.

I could certainly add code that populates `/etc/hosts` accordingly for this case. However this also adds code that needs to be considered when debugging.

smithfarm:
Given that

1. we are already approaching the overall issue by populating `/etc/hosts` in the default case
2. `--libvirt-networks` is currently skipping this step

could we agree that `--libvirt-networks` not populating `/etc/hosts` is the root cause of the bug?

If that is the case, then the correct approach IMO would be to populate `/etc/hosts` in the `--libvirt-networks` case. Then it would not be necessary to change the behavior of `hostname`, set the `domain` grain, etc.

jan--f:
ok @smithfarm how about this version here? I think this does everything we need.

jan--f:
hmm the lint task is happy locally...

smithfarm:
> hmm the lint task is happy locally...

Did you try `tox -e lint` ?

smithfarm:
Ah, this is something I introduced - it's being fixed now and can be ignored:

```
sesdev/__init__.py:129:4: W1201: Specify string format arguments as logging function parameters (logging-not-lazy)
```

jan--f:
@smithfarm can the CI run be retriggered? Seems like there was an issue with http requests to github.

smithfarm:
> @smithfarm can the CI run be retriggered? Seems like there was an issue with http requests to github.

I don't know. Question for @toabctl maybe?

smithfarm:
I tested this manually, and there appears to be a problem with it, which I do not currently understand.

If a time server is set, ceph-salt will install the `chrony` package on all minions.

However, when this PR is active, ceph-salt does not install the `chrony` package on the `master` minion. Only on non-master minions.

jan--f:
That seems like a bug in ceph-salt. Which command did you test there? tried an octopus cluster but that ends in a python stack trace.

smithfarm:
Yeah, I'm seeing:

```
    master: ++ ceph-salt config /Ceph_Cluster/Minions add node3.octopus-4node.com
    master: Host 'node3.octopus-4node.com' FQDN resolves to the loopback interface IP address
    master: ++ ceph-salt config /Ceph_Cluster/Roles/Admin add node3.octopus-4node.com
    master: No minions matched "node3.octopus-4node.com".
```

smithfarm:
The command I used to test:

```
sesdev create octopus --ceph-salt-repo https://github.com/ceph/ceph-salt --ceph-salt-branch master --qa-test octopus-4node
```

jan--f:
hmm we hit https://github.com/ceph/ceph-salt/issues/160 with this.

toabctl:
test this please

toabctl:
test this please

toabctl:
test this please

jan--f:
ping @smithfarm is this still failing any qa tests?

smithfarm:
@jan--f I'll include it in my next `contrib/standalone.sh` run.

smithfarm:
@jan--f Could you rebase this on top of latest master and then try the following command?

    sesdev create octopus --single-node --non-interactive

I'm getting:

```
(venv) smithfarm@geras:~/sesdev> sesdev create octopus --single-node --non-interactive
=== Creating deployment "octopus_mini" with the following configuration ===
1 Deployment VMs:
  -- master:
     - OS:               leap-15.2
     - ses_version:      octopus
     - deployment_tool:  cephadm
     - roles:            ['bootstrap', 'grafana', 'igw', 'master', 'mds', 'mgr', 'mon', 'nfs', 'prometheus', 'rgw', 'storage']
                         (CAVEAT: the 'admin' role is assumed even though not explicitly given!)
     - fqdn:             master.octopus_mini.com
     - public_address:   10.20.65.200
     - cpus:             2
     - ram:              4G
     - storage_disks:    4
                         (device names will be assigned by vagrant-libvirt)
     - encrypted OSDs:   False
     - repo_priority:    True
     - qa_test:          False
     - image_path:       registry.opensuse.org/filesystems/ceph/octopus/upstream/images/ceph/ceph


=== Running shell command ===
vagrant up
Bringing machine 'master' up with 'libvirt' provider...
There are errors in the configuration of this machine. Please fix
the following errors and try again:

vm:
* The hostname set for the VM 'master' should only contain letters, numbers,
hyphens or dots. It cannot start with a hyphen or dot.
```

smithfarm:
Note: the 

```
vm:
* The hostname set for the VM 'master' should only contain letters, numbers,
hyphens or dots. It cannot start with a hyphen or dot.
```

issue was fixed by https://github.com/SUSE/sesdev/pull/335

jan--f:
Thanks @smithfarm !

PR #202, opened 2020-03-27
sesdev: fail when master role is missing

smithfarm:
Without a master role, the deployment will never succeed, so don't even
start it.

### DIFF
--- a/seslib/exceptions.py
+++ b/seslib/exceptions.py
@@ -121,3 +121,11 @@ class SupportconfigOnlyOnSLE(SesDevException):
             "sesdev supportconfig depends on the 'supportconfig' RPM, which is "
             "available only on SUSE Linux Enterprise"
             )
+
+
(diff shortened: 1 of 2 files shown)


seslib/exceptions.py
```diff
@@ -121,3 +121,10 @@ def __init__(self):
             "sesdev supportconfig depends on the 'supportconfig' RPM, which is "
             "available only on SUSE Linux Enterprise"
             )
+
+class NoMasterRoleInRoles(SesDevException):
+    def __init__(self):
+        super(NoMasterRoleInRoles, self).__init__(
+            "No 'master' role specified - please check the roles and make sure one, "
```
toabctl: do we already check that there can't be more than one somehere?

smithfarm: good idea, I'll add one

PR #204, opened 2020-03-27
seslib: raise exception if more, or less, than one master/bootstrap role

smithfarm:
A check that quickly bails out if someone omits the "master" role was
implemented by ddd4d91 but that commit
is being reverted and replaced by this solution which implements the
check in a less intrusive way, later in the "sesdev create" process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,6 +287,7 @@ nodes are separated by commas, too.
 The following roles can be assigned:
 
 * `master` - The master node, running management components like the Salt master
+* `bootstrap` - The node where `cephadm bootstrap` will be run
 * `client` - Various Ceph client utilities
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -300,12 +301,13 @@ The following roles can be assigned:
 * `suma` - SUSE Manager (octopus only)
 
 The following example will generate a cluster with four nodes: the master (Salt
-Master) node that is also running a MON daemon, two storage (OSD) nodes that
-will also run a MON, a MGR and an MDS, and another node that will run an iSCSI
+Master) node that is also running a MON daemon, a storage (OSD) nodes that
```
ricardoasmarques: s/nodes/node/

PR #205, opened 2020-03-27
deepsea_deployment: run Stage 4 only if justified by roles

smithfarm:
Stage 4 is not needed, and is a noop, on basic "MON, MGR, OSD" clusters, so omit it in such cases.

### DIFF
--- a/seslib/templates/deepsea/deepsea_deployment.sh.j2
+++ b/seslib/templates/deepsea/deepsea_deployment.sh.j2
@@ -136,6 +136,7 @@ ceph config set global mon_max_pg_per_osd 500
 exit 0
 {% endif %}
 
+{% if deepsea_need_stage_4 %}
 echo ""
 echo "***** RUNNING stage.4 *******"
 {% if use_deepsea_cli %}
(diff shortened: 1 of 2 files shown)


smithfarm:
Tested with: 

```
sesdev create nautilus --roles="[ master, storage, mon, mgr ]" nautilus
```

Works as advertised.

seslib/templates/deepsea/deepsea_deployment.sh.j2
```diff
@@ -136,14 +136,15 @@ ceph config set global mon_max_pg_per_osd 500
...
 echo "***** RUNNING stage.4 *******"
 {% if use_deepsea_cli %}
 deepsea stage run --simple-output ceph.stage.4
 {% else %}
 salt-run state.orch ceph.stage.4
 {% endif %}
-sleep 5
+{% endif %} {# deepsea_need_stage_4 #}
 
+sleep 5
```
tserong, line 149: Should that `sleep` be inside the deepsea_need_stage_4 block?

smithfarm: yes indeed.

Issue #206, opened 2020-03-31
sesdev should not use ".ssh/id_rsa" and ".ssh/id_rsa.pub"


smithfarm:
The files `.ssh/id_rsa` and `.ssh/id_rsa.pub` might be used by software that is deployed by sesdev (e.g. Cephadm). The fact that sesdev generates this keypair and installs it on all the VMs can hide bugs in the software under test.

PR #207, opened 2020-03-31
seslib: install sesdev-generated keypair under non-default name

smithfarm:
For each deployment, sesdev generates an SSH keypair and:

1. uploads it as /home/vagrant/.ssh/id_rsa{.pub} to all the VMs
2. copies it to /root/.ssh/id_rsa{.pub}
3. appends the public part to /home/vagrant/.ssh/authorized_keys
4. appends the public part of /root/.ssh/authorized_keys

This is problematic because /root/.ssh/id_rsa{.pub} is prone to being
rewritten by "other software" (e.g. ceph-salt was rewriting it).

This patch makes sesdev install its keypair under a different name:
".ssh/sesdev{.pub}".

To avoid losing the ability to ssh from any node of the cluster to any
other node simply by issuing the command "ssh $NODE_ID" as root, we
create "/root/.ssh/id_rsa{.pub}" as a symlink to sesdev's keypair.

Fixes: https://github.com/SUSE/sesdev/issues/206

### DIFF
--- a/seslib/templates/provision.sh.j2
+++ b/seslib/templates/provision.sh.j2
@@ -87,11 +87,13 @@ zypper mr -p {{ repo.priority }} {{ repo.name }}
 zypper --gpg-auto-import-keys ref
 {% endif %}
 
-cat /home/vagrant/.ssh/id_rsa.pub >> /home/vagrant/.ssh/authorized_keys
+cat /home/vagrant/.ssh/{{ ssh_key_name }}.pub >> /home/vagrant/.ssh/authorized_keys
 [ ! -e "/root/.ssh" ] && mkdir /root/.ssh
-chmod 600 /home/vagrant/.ssh/id_rsa
-cp /home/vagrant/.ssh/id_rsa* /root/.ssh/
-cat /root/.ssh/id_rsa.pub >> /root/.ssh/authorized_keys
+chmod 600 /home/vagrant/.ssh/{{ ssh_key_name }}
+cp /home/vagrant/.ssh/{{ ssh_key_name }}* /root/.ssh/
+ln -s /root/.ssh/{{ ssh_key_name }} /root/.ssh/id_rsa
+ln -s /root/.ssh/{{ ssh_key_name }}.pub /root/.ssh/id_rsa.pub
+cat /root/.ssh/{{ ssh_key_name }}.pub >> /root/.ssh/authorized_keys
 hostnamectl set-hostname {{ node.name }}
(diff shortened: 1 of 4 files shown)


smithfarm:
I have tested this and could not find anything broken by it.

seslib/templates/provision.sh.j2
```diff
@@ -87,11 +87,13 @@ zypper mr -p {{ repo.priority }} {{ repo.name }}
...
 
-cat /home/vagrant/.ssh/id_rsa.pub >> /home/vagrant/.ssh/authorized_keys
+cat /home/vagrant/.ssh/{{ ssh_key_name }}.pub >> /home/vagrant/.ssh/authorized_keys
 [ ! -e "/root/.ssh" ] && mkdir /root/.ssh
-chmod 600 /home/vagrant/.ssh/id_rsa
-cp /home/vagrant/.ssh/id_rsa* /root/.ssh/
-cat /root/.ssh/id_rsa.pub >> /root/.ssh/authorized_keys
+chmod 600 /home/vagrant/.ssh/{{ ssh_key_name }}
+cp /home/vagrant/.ssh/{{ ssh_key_name }}* /root/.ssh/
+ln -s /root/.ssh/{{ ssh_key_name }} /root/.ssh/id_rsa
```
toabctl, line 94: Isn't that symlink then also overwritten by other software? Shouldn't we instead use `ssh -i {{ ssh_keyname }}` everywhere so sesdev is always using the key it created?

smithfarm: > Shouldn't we instead use ssh -i {{ ssh_keyname }} everywhere so sesdev is always using the key it created?

This is already the case. Sesdev only sshes into the cluster from the host, and it stores the private key in a non-default location on the host (`~/.sesdev/${DEP_ID}/keys`). This PR is about the keypair that sesdev installs on the cluster nodes, which is there only as a convenience to the user. Sesdev itself should not use this keypair, though I have not studied all the scripts under `seslib/templates/...` to be completely certain this is the case.

> Isn't that symlink then also overwritten by other software?

Yes, it might be. The point here is that, when the "other software" overwrites `/root/.ssh/id_rsa{.pub}`, before this patch it was overwriting the actual keypair that sesdev installed on the nodes. Now it overwrites the symlink only, so the actual keypair is not lost when that happens.

toabctl: ok. Thanks!

PR #210, opened 2020-04-02
ceph_salt_deployment: make it easier to install ceph-salt from source

smithfarm:
Before, one had to provide both "--ceph-salt-repo" and
"--ceph-salt-branch" in order for sesdev to install ceph-salt from
source.

With this patch, only one or the other is required. When one of these
two options is given, the other will default to a sensible value.

The defaults are hard-coded to:

--ceph-salt-repo https://github.com/ceph/ceph-salt
--ceph-salt-branch master


tserong:
Needs rebase

PR #211, opened 2020-04-02
Require bootstrap role only for SES7/Octopus

colder-is-better:
Checking for the presence of role 'bootstrap' makes sense for SES7/Octopus clusters only.


smithfarm:
@colder-is-better Sorry, I didn't see this when I opened #213 

Can you make yours look like #213 ? So it passes the linter. Also, can you sign your commit (`git commit --signoff`)?

Another possibility is you could review #213 and I'll merge that one :-)

PR #214, opened 2020-04-02
Jenkinsfile.integration: Retry jenkins slave deletion

toabctl:
In case the node delete fails, retry it. We don't want to have running
nodes on OVH after a job finished.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -167,7 +167,10 @@ pipeline {
             node("${jenkins_slave_base}") {
                 sh "virtualenv venv"
                 sh "source venv/bin/activate; pip install git+https://github.com/toabctl/jcs.git#egg=jcs[openstack,obs,jenkins]"
-                sh "source venv/bin/activate; jcs delete --cloud openstack ${jenkins_slave_name}"
+                retry(10) {
+                    sh "source venv/bin/activate; jcs delete --cloud openstack ${jenkins_slave_name}"
+                    sleep 10
+                }
             }
         }
     }


toabctl:
See eg. https://ceph-ci.suse.de/blue/organizations/jenkins/sesdev-integration/detail/PR-199/3/pipeline which failed.

Jenkinsfile.integration
```diff
@@ -167,7 +167,9 @@ pipeline {
             node("${jenkins_slave_base}") {
                 sh "virtualenv venv"
                 sh "source venv/bin/activate; pip install git+https://github.com/toabctl/jcs.git#egg=jcs[openstack,obs,jenkins]"
```
kshtsk, line 169: Why do you need to do this all the time? it is enough to this once in the beginning of the job, no?

Jenkinsfile.integration
```diff
@@ -167,7 +167,9 @@ pipeline {
...
-                sh "source venv/bin/activate; jcs delete --cloud openstack ${jenkins_slave_name}"
+                retry(10) {
+                    sh "source venv/bin/activate; jcs delete --cloud openstack ${jenkins_slave_name}"
```
kshtsk, line 171: Maybe add a pause?

Jenkinsfile.integration
toabctl: No. A stage start with a clean state. We would need to use `stash` and `unstash` to reuse stuff from other stages.

Jenkinsfile.integration
toabctl: Done

Jenkinsfile.integration
```diff
@@ -167,7 +167,10 @@ pipeline {
...
+                    sleep 10
```
kshtsk, line 172: Instead of doing:
```failure {
            node("${jenkins_slave_base}") {
                sh "sleep ${params.SLEEP_WHEN_FAILING.toInteger() * 60}"
            }
        }
```
Prior cleanup, you can add this sleep before trying to run `jcs delete`

Jenkinsfile.integration
kshtsk: Why? It is not required, you should be able to reuse the workspace in any stage.

toabctl: feel free to try it...

Jenkinsfile.integration
toabctl: ? I want to sleep in the retry loop, no?

Jenkinsfile.integration
toabctl: but unrelated to this commit so resolving the conversation for now. Please open a PR if you think this needs fixing.

Jenkinsfile.integration
kshtsk: I mean, why do you need `failure` block?
```
    post {
        // cleanup is executed after every other post step
        failure {
            node("${jenkins_slave_base}") {
                sh "sleep ${params.SLEEP_WHEN_FAILING.toInteger() * 60}"
            }
        }
        cleanup {
            node("${jenkins_slave_base}") {
                sh "virtualenv venv"
                sh "source venv/bin/activate; pip install git+https://github.com/toabctl/jcs.git#egg=jcs[openstack,obs,jenkins]"
                sh "source venv/bin/activate; jcs delete --cloud openstack ${jenkins_slave_name}"
            }
        }
    }
```
So I suggest to move it to the loop like this:
```
    post {
        // cleanup is executed after every other post step
        cleanup {
            node("${jenkins_slave_base}") {
                sleep 10
                sh """# cleanup node
                      virtualenv venv
                      source venv/bin/activate
                      pip install git+https://github.com/toabctl/jcs.git#egg=jcs[openstack,obs,jenkins]"
                      jcs delete --cloud openstack ${jenkins_slave_name}"""
            }
        }
    }
```

toabctl: ? how is that related to this PR? Where is the `retry`?

toabctl:
@kshtsk please create a PR with your suggestion to fix it.

Issue #216, opened 2020-04-03
ses7 deployment fails with: AttributeError: 'NoneType' object has no attribute 'startswith' for self.settings.ceph_salt_git_branch


toabctl:
On TW, with git commit 022cedfef463ef3f7ea6d7fa9bccc62385b00de5

```
$ sesdev create ses7 --single-node ses7-mini
Traceback (most recent call last):
  File "/home/tom/devel/suse/sesdev/venv/bin/sesdev", line 11, in <module>
    load_entry_point('sesdev', 'console_scripts', 'sesdev')()
  File "/home/tom/devel/suse/sesdev/sesdev/__init__.py", line 17, in sesdev_main
    cli(prog_name='sesdev')
  File "/home/tom/devel/suse/sesdev/venv/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/tom/devel/suse/sesdev/venv/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/tom/devel/suse/sesdev/venv/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/tom/devel/suse/sesdev/venv/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/tom/devel/suse/sesdev/venv/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/tom/devel/suse/sesdev/venv/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/tom/devel/suse/sesdev/sesdev/__init__.py", line 660, in ses7
    _create_command(deployment_id, deploy, settings_dict)
  File "/home/tom/devel/suse/sesdev/sesdev/__init__.py", line 576, in _create_command
    dep = seslib.Deployment.create(deployment_id, settings)
  File "/home/tom/devel/suse/sesdev/seslib/__init__.py", line 1590, in create
    dep.save()
  File "/home/tom/devel/suse/sesdev/seslib/__init__.py", line 1163, in save
    scripts = self.generate_vagrantfile()
[...]

PR #217, opened 2020-04-03
seslib: make all str-type settings default to the empty string

smithfarm:
When string variables are set to None, we can easily fall into a trap
like this:

```
  File "/home/tom/devel/suse/sesdev/seslib/__init__.py", line 1082, in generate_vagrantfile
      if self.settings.ceph_salt_git_branch.startswith('origin/pr/'):
  AttributeError: 'NoneType' object has no attribute 'startswith'
```

Fixes: https://github.com/SUSE/sesdev/issues/216

### DIFF
--- a/seslib/__init__.py
+++ b/seslib/__init__.py
@@ -233,7 +233,7 @@ SETTINGS = {
     'os': {
         'type': str,
         'help': 'openSUSE OS version (leap-15.1, tumbleweed, sles-12-sp3, or sles-15-sp1)',
-        'default': None,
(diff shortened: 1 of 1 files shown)


toabctl:
I like https://github.com/SUSE/sesdev/pull/218 more.

smithfarm:
> I like #218 more.

Cool. We can have both :-)

smithfarm:
We have list-type settings defaulting to the empty list, so having str-type settings default to the empty string is more consistent.

PR #223, opened 2020-04-06
libvirt/vagrant.provider.j2: Use less VRAM for VMs

colder-is-better:
By default Vagrant creates libvirt VMs with 9MB of Video RAM.
We don't need that much memory for headless VMs, so we're
instructing libvirt to create VMs with just 2MB of VRAM.

Signed-off-by: Christos Varelas <cvarelas@suse.de>

### DIFF
--- a/seslib/templates/engine/libvirt/vagrant.provider.j2
+++ b/seslib/templates/engine/libvirt/vagrant.provider.j2
@@ -15,6 +15,7 @@
 {% if libvirt_storage_pool %}
     lv.storage_pool_name = "{{ libvirt_storage_pool }}"
 {% endif %}
+    lv.video_vram = 2048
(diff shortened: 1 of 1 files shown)


toabctl:
out of interest - do you run into any problems with the vagrant default?

colder-is-better:
No, there's no problem with the default setting of 9MB of VRAM.

smithfarm:
I'm a little uncomfortable changing this when the default is not causing any problems for anyone.

Maybe "If it ain't broke, don't fix it" applies here?

colder-is-better:
Well, in my Vagrant-based labs I never had any issue because of this (i.e., a VRAM size lower than the default). On the other hand, I totally get the "If it ain't broke, don't fix it" mantra.

PR #227, opened 2020-04-07
ceph_salt_deployment.sh: do not refresh/sync pillar data

smithfarm:
As of https://github.com/ceph/ceph-salt/pull/157 ceph-salt should take
care of this itself.

Also, since these steps might not be included in the documentation, it's
possible that a user will omit them and run into bugs that, because of
these two lines, are not seen with sesdev.


smithfarm:
passed a full `contrib/standalone.sh` run

PR #230, opened 2020-04-07
templates: extend --qa-test to DeepSea-deployed versions and fix issues preventing it from passing

smithfarm:
Until now, `--qa-test` and `sesdev qa-test` have only worked when ceph-salt is used for deployment, but there is no good reason for that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ Pipfile
 build
 dist
 .tox
+typescript
--- a/contrib/run-standalone.sh
+++ b/contrib/run-standalone.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+#
+# run sesdev test script "contrib/standalone.sh" while leveraging
+# GNU script to capture all output in a file called "typescript"
+#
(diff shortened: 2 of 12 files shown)


smithfarm:
I am removing the `needs-qa` label, because the new `contrib/standalone.sh` test passes on this PR with the following invocation:

    contrib/run-standalone.sh --ceph-salt-from-source

(The `--ceph-salt-from-source` option is needed because of https://github.com/ceph/ceph-salt/pull/166.)

`contrib/standalone.sh` is a script which

* deploys 1- and 4-node clusters of each of the main deployment types (`ses5, `nautilus`, `ses6`, `octopus`, `ses7`, and `pacific`)
* runs basic QA tests on each cluster after deployment is finished
* if QA tests succeed, cluster is destroyed before moving on to the next one
* in case of a failure, the script aborts to enable a post-mortem to be performed

`contrib/standalone.sh` also takes options enabling the scope of testing to be reduced. For example, one can do:

    contrib/standalone.sh --ceph-salt-from-source --octopus --ses7

PR #234, opened 2020-04-07
templates: move test.ping try-wait into a separate file

smithfarm:
The exact same code was used in multiple places. Take advantage of
Jinja's "include" directive to have it in only one place.


smithfarm:

PR #235, opened 2020-04-07
sesdev: make deployment_id argument optional

smithfarm:
When one is accustomed to creating and destroying lots of clusters,
it can be tiresome to type in the deployment_id argument, always the
same one, for each invocation.

Make the argument optional and have it default to something sane.

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,8 +288,9 @@ system, run the following command:
 $ sesdev create nautilus --single-node mini
 ```
 
-The `mini` argument is the ID of the deployment. You can create many deployments
-by giving them different IDs.
+The `mini` argument is the ID of the deployment. It is optional: if you omit it,
+sesdev will assign an ID as it sees fit. You can create many deployments by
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -285,11 +285,12 @@ To create a single node Ceph cluster based on nautilus/leap-15.1 on your local
 system, run the following command:
 
 ```
-$ sesdev create nautilus --single-node mini
+$ sesdev create nautilus --single-node
 ```
 
-The `mini` argument is the ID of the deployment. You can create many deployments
-by giving them different IDs.
+The `mini` argument is the ID of the deployment. It is optional: if you omit it,
```
ricardoasmarques, line 291: If we are referring to the `mini` argument here, then we should not remove it from the command above.

smithfarm: good catch

smithfarm:
passed a full `contrib/standalone.sh` run

Issue #82, opened 2020-02-04
feature request: "sesdev make-check TARGET"


smithfarm:
The overall CI effort includes regularly running "make check" for the various products/environments (e.g. SES5 on SLE-15-SP3, Octopus on openSUSE Leap 15.2, etc.)

Implement a "sesdev make-check TARGET" command which would:

* deploy a single node with 16 GB of memory
* add all the repos needed to run "make check" on that target
* install sudo package
* create a normal user and give it passwordless sudo privileges
* as that normal user:
    * clone a ceph repo/branch (each target would have a default repo/branch, which could be overrided by supplying `--repo` and/or `--branch` options)
    * run `install-deps.sh`, and if that succeeds:
    * run `run-make-check.sh`

toabctl:
I wonder how often we want to reimplement the `run-make-check` setup now? There is:
- upstream on openSUSE Leap which does run `run-make-check`
- downstream for every distro we support a jenkins job that does `run-make-check`
- now this request

Why do we want/need it here again? I'm fine with doing it, but then I would like to remove our other downstream approach with running `run-make-check`. I think we (as a team) should concentrate on one approach for runing `run-make-check` and improve and use that approach...

smithfarm:
@toabctl The Jenkins job could be modified to use `sesdev make-check` instead of what it does now :-)

The motivation for implementing it in sesdev is to provide developers another tool for reproducing intermittent make check failures that are seen in Jenkins. (For example, Jenkins runs the tests in a VM, and sesdev runs its workloads in a VM, and some test failures might occur only, or more frequently, in VMs.)

Each tool has its strengths and weaknesses. Jenkins is not so great when it comes to reproducing intermittent failures and letting a developer SSH into the environment and do a post-mortem.

PR #236, opened 2020-04-07
Implement feature "sesdev create makecheck"

smithfarm:
When the command is issued, sesdev will create and provision a single
VM, and inside that VM it will create an ordinary user and grant it
passwordless sudo privileges. That user will then clone the git repo,
check out the branch, and run install-deps.sh and run-make-check.sh
inside it.

Originally, this feature was envisioned as "sesdev make-check VERSION",
but during the course of implementation the following was found to be
more expedient:

    sesdev create makecheck \
        --os leap-15.2 \
        --ceph-repo https://github.com/ceph/ceph \
        --ceph-branch octopus

(The main reason is that "make check" is tied more to operating system
than to SES version.)

Fixes: https://github.com/SUSE/sesdev/issues/82

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ the VMs and run the deployment scripts.
    * [Install sesdev from source](#install-sesdev-from-source)
       * [Running the unit tests](#running-the-unit-tests)
 * [Usage](#usage)
-   * [Create/deploy a cluster](#createdeploy-a-cluster)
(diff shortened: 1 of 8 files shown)


smithfarm:

theanalyst:
This one works on my end so +1 from me

smithfarm:
This passed another full `contrib/standalone.sh` run.

Issue #242, opened 2020-04-13
sles-12-sp3 Vagrant Box has too-new versions of python-base, python-xml, libncurses5, and libncurses6


smithfarm:
The SLE-12-SP3 Vagrant Box has newer versions of the python-base, python-xml, libncurses5, and libncurses6 packages than are available in the standard repos. This is a Bad Thing and, I might add, a bug in the SLE-12-SP3 Vagrant Box.

This can be demonstrated by running `zypper se -s` on the packages in question. The newest version is coming from `(System Packages)`, which in our case means it is on the underlying image.

The newest python-base package is in "storage-update" because the SES5 Update channel gets SLE-12-SP3 LTSS updates:

```
master:~ # zypper se -s python-base
Loading repository data...
Reading installed packages...

S | Name                | Type       | Version        | Arch   | Repository       
--+---------------------+------------+----------------+--------+------------------
  | openlmi-python-base | package    | 0.4.2-6.23     | noarch | base             
i | python-base         | package    | 2.7.13-28.36.1 | x86_64 | (System Packages)
v | python-base         | package    | 2.7.13-28.31.1 | x86_64 | storage-update   
v | python-base         | package    | 2.7.13-28.26.1 | x86_64 | update           
v | python-base         | package    | 2.7.13-28.23.1 | x86_64 | update           
v | python-base         | package    | 2.7.13-28.21.1 | x86_64 | update           
v | python-base         | package    | 2.7.13-28.16.1 | x86_64 | update           
v | python-base         | package    | 2.7.13-28.6.1  | x86_64 | update           
v | python-base         | package    | 2.7.13-28.3.2  | x86_64 | update           
v | python-base         | package    | 2.7.13-27.1    | x86_64 | base             
```

The latest versions of libncurses5 and libncurses6 are in the ordinary SLE-12-SP3 Update channel:

```
master:~ # zypper se -s libncurses5
Loading repository data...
Reading installed packages...

S | Name              | Type    | Version    | Arch   | Repository       
--+-------------------+---------+------------+--------+------------------
i | libncurses5       | package | 5.9-72.1   | x86_64 | (System Packages)
v | libncurses5       | package | 5.9-64.1   | x86_64 | update           
v | libncurses5       | package | 5.9-61.1   | x86_64 | update           
v | libncurses5       | package | 5.9-58.1   | x86_64 | update           
v | libncurses5       | package | 5.9-55.1   | x86_64 | update           
[...]

PR #243, opened 2020-04-13
provision.sh: reinstall certain packages from the update repos

smithfarm:
The SLE-12-SP3 Vagrant Box has newer versions of the python-base,
python-xml, libncurses5, and libncurses6 packages than are available in
the standard repos. This is a Bad Thing and, I might add, a bug in the
SLE-12-SP3 Vagrant Box. It also makes install-deps.sh fail. Work around
the bug by reverting these four packages to the latest version that is
in the repos.

Fixes: https://github.com/SUSE/sesdev/issues/242

### DIFF
--- a/seslib/templates/provision.sh.j2
+++ b/seslib/templates/provision.sh.j2
@@ -48,7 +48,12 @@ zypper addrepo --refresh
 {% endif %}
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


smithfarm:

Issue #245, opened 2020-04-13
sesdev create chokes on node with no roles


smithfarm:
The following command causes sesdev to exhibit buggy behavior:

```
$ sesdev create octopus --ceph-salt-branch master \
    --roles [[master,mon,mgr,storage,bootstrap,client],[]] \
    octopus-2node
```

Parsing chaos ensues:

```
=== Creating deployment with the following configuration ===
2 Deployment VMs:
  -- node1:
     - OS:               leap-15.2
     - ses_version:      octopus
     - roles:            ['storage', '[master', 'client', 'bootstrap', 'mgr', 'mon']
                         (CAVEAT: the 'admin' role is assumed even though not explicitly given!)
     - fqdn:             node1.octopus-2node.com
     - public_address:   10.20.75.201
     - cpus:             2
     - ram:              4G
     - storage_disks:    4
                         (device names will be assigned by vagrant-libvirt)
     - encrypted OSDs:   False
     - repo_priority:    True
     - qa_test:          False
     - image_path:       registry.opensuse.org/filesystems/ceph/octopus/upstream/images/ceph/ceph

  -- node2:
     - OS:               leap-15.2
     - ses_version:      octopus
     - roles:            [']']
                         (CAVEAT: the 'admin' role is assumed even though not explicitly given!)
     - fqdn:             node2.octopus-2node.com
     - public_address:   10.20.75.202
     - cpus:             2
     - ram:              4G
     - repo_priority:    True
     - qa_test:          False
     - image_path:       registry.opensuse.org/filesystems/ceph/octopus/upstream/images/ceph/ceph


Do you want to continue with the deployment? [Y/n]: 
```

smithfarm:
This is a good opportunity to implement some unit tests for `_parse_roles`

PR #246, opened 2020-04-13
ceph_salt_deployment: do not provision client-only nodes

smithfarm:
ceph_salt_deployment.sh should not ask ceph-salt to provision:

- nodes with no roles at all
- nodes that have the "client" role and *only* the "client" role

because these nodes might be used later for QA testing.

---

seslib: refuse to process unknown roles

Also implement Node class helper methods "has_roles" and
"has_exclusive_role".

---

sesdev/_parse_roles: handle node with no roles

Fixes: https://github.com/SUSE/sesdev/issues/245

---

tests/test_sesdev_parse_roles.py: unit tests for _parse_roles

### DIFF
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -447,10 +447,14 @@ def _gen_settings_dict(version,
     if not single_node and roles:
         settings_dict['roles'] = _parse_roles(roles)
     elif single_node:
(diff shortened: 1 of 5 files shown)


smithfarm:

Issue #237, opened 2020-04-08
allow user-defined nfs mounts provided via the cli


jschmid1:
Vagrant allows nfs shares to be mounted inside a VM (https://www.vagrantup.com/docs/synced-folders/nfs.html).

sesdev doesn't use this feature currently.
https://github.com/SUSE/sesdev/blob/78f238a80861b2e76cb7f82287a6c80281ae242e/seslib/templates/Vagrantfile.j2#L28

Having a option to pass an arbitrary directory that will get mounted on the master is going to be helpful for cephadm/ceph-volume devlopment (or testing in general).
This allows you to change files on your local file system with your favorite editor and have them in sync on the VM you're working on. It also avoids copying directories back and forth.

Maybe something like this:

`sesdev .... --sync-folder|--nfs-mount <path_to_dir=path_to_mount>`
 
the separator needs some discussion. '=', '-', ':'. Podman uses a colon for this, so maybe we should stay with this convention

smithfarm:
This would be a fine feature to have, I think.

Given that the vagrant option is `synced_folder`, would it make sense to call the option `--synced-folder` ?

smithfarm:
And, I might add, adhering to Podman's `:` convention sounds reasonable.

smithfarm:
Question: will multiple instances of the option be supported (for syncing multiple folders), or just one?

jschmid1:
> Given that the vagrant option is `synced_folder`, would it make sense to call the sesdev option `--synced-folder` ?

Let's go with `synced-folder`


> Question: will multiple instances of the option be supported (for syncing multiple folders), or just one?

I'd say so.  Either

`--synced-folder src1:dst1 src2:dst2`

or

`--synced-folder src1:dst1 --synced-folder src2:dst2`

Depending on what the argparser will accept more easily I think.



> And, I might add, adhering to the Docker/Podman (`:`) convention sounds reasonable.

`:` it is then

smithfarm:
> `--synced-folder src1:dst1 src2:dst2`

This would have to be `--synced-folder "src1:dst1 src2:dst2"` if you wanted to go this route - otherwise Click would interpret the second pair as an argument (positional).

> `--synced-folder src1:dst1 --synced-folder src2:dst2`

This one gets my vote.

PR #247, opened 2020-04-14
Implement feature: '--synced-folder' to nfs mount directories

jschmid1:
resolves https://github.com/SUSE/sesdev/issues/237

example ouput
```
➜  sesdev create octopus --synced-folder /home/jxs/projects/ceph:/root/ceph  test_sync --single-node
=== Creating deployment "test_sync" with the following configuration ===
1 Deployment VMs:
  -- master:
     - OS:               leap-15.2
[..snip..]
     - synced_folder:    /home/jxs/projects/ceph -> /root/ceph
```

### DIFF
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -1,10 +1,11 @@
 import logging
 import sys
+from os.path import isabs, exists, isdir
 
 import click
 import pkg_resources
 import seslib
-from seslib.exceptions import SesDevException
+from seslib.exceptions import SesDevException, OptionFormatError, OptionValueError
 
 
 logger = logging.getLogger(__name__)
@@ -119,6 +120,8 @@ def common_create_options(func):
                      help='Do not ask the user if they really want to'),
         click.option('--encrypted-osds', is_flag=True, default=False,
                      help='Deploy OSDs encrypted'),
+        click.option('--synced-folder', type=str, default=None, multiple=True,
+                     help='Set synced-folder to be mounted on the master node. <str:dest>'),
     ]
     return _decorator_composer(click_options, func)
 
@@ -426,6 +429,7 @@ def _gen_settings_dict(version,
                        scc_pass,
                        domain,
                        non_interactive,
+                       synced_folder,
                        encrypted_osds,
                        deepsea_cli=None,
                        stop_before_deepsea_stage=None,
@@ -588,6 +592,23 @@ def _gen_settings_dict(version,
 
     if not ceph_salt_deploy:
         settings_dict['ceph_salt_deploy'] = False
+
+    for folder in synced_folder:
+        try:
+            src, dst = folder.split(':')
+            if not all([isabs(x) for x in [src, dst]]):
+                raise OptionValueError('--synced-folder',
+                                       "Please provide absolute paths for "
+                                       "synced folder paths",
+                                       folder)
+            if not exists(src):
+                raise OptionValueError('--synced-folder',
+                                       "Path to the source synced folder must exist",
+                                       src)
+
+        except ValueError:
+            raise OptionFormatError('--synced-folder', "src:dst", folder)
+    settings_dict['synced_folder'] = [folder.split(':') for folder in synced_folder]
 
     return settings_dict
 
--- a/seslib/__init__.py
+++ b/seslib/__init__.py
@@ -482,6 +482,11 @@ SETTINGS = {
         'help': 'Deploy SES using rook in CaasP',
         'default': False,
     },
+    'synced_folder': {
+        'type': list,
+        'help': 'Sync Folders to VM',
+        'default': [],
+    },
 }
 
 
@@ -1149,6 +1154,7 @@ class Deployment():
             'ceph_salt_deploy': self.settings.ceph_salt_deploy,
             'node_manager': NodeManager(list(self.nodes.values())),
             'caasp_deploy_ses': self.settings.caasp_deploy_ses,
+            'synced_folders': self.settings.synced_folder,
         }
 
         scripts = {}
@@ -1372,6 +1378,8 @@ class Deployment():
             if self.settings.version in ['octopus', 'ses7', 'master'] \
                     and self.settings.deployment_tool == 'cephadm':
                 result += "     - image_path:       {}\n".format(self.settings.image_path)
+            for synced_folder in self.settings.synced_folder:
+                result += "     - synced_folder:    {}\n".format(' -> '.join(synced_folder))
             if v.repos:
                 result += "     - custom_repos:\n"
                 for repo in v.repos:
--- a/seslib/exceptions.py
+++ b/seslib/exceptions.py
@@ -45,6 +45,20 @@ class SettingTypeError(SesDevException):
             .format(setting, expected_type, value, type(value)))
 
 
+class OptionValueError(SesDevException):
+    def __init__(self, option, message, value):
+        super(OptionValueError, self).__init__(
+            "Wrong value for option '{}'. {}. Actual value: '{}'"
+            .format(option, message, value))
+
+
+class OptionFormatError(SesDevException):
+    def __init__(self, option, expected_type, value):
+        super(OptionFormatError, self).__init__(
+            "Wrong format for option '{}': expected format: '{}', actual format: '{}'"
+            .format(option, expected_type, value))
+
+
 class VagrantBoxDoesNotExist(SesDevException):
     def __init__(self, box):
         super(VagrantBoxDoesNotExist, self).__init__(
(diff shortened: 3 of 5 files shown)


jschmid1:
fixing:
```
sesdev/__init__.py:584:8: C0103: Variable name "sf" doesn't conform to snake_case naming style (invalid-name)
sesdev/__init__.py:8:0: C0411: standard import "from os.path import isabs, exists, isdir" should be placed before "import click" (wrong-import-order)
```

seslib/exceptions.py
```diff
@@ -45,6 +45,20 @@ def __init__(self, setting, expected_type, value):
...
+    def __init__(self, setting, message, value):
+        super(SettingValueError, self).__init__(
+            "Wrong value for setting '{}'. {}. Actual value: '{}'"
+            .format(setting, message, value))
+
+
+class SettingFormatError(SesDevException):
+    def __init__(self, setting, expected_type, value):
+        super(SettingFormatError, self).__init__(
+            "Wrong format for setting '{}': expected value: '{}', actual value: '{}'"
```
smithfarm, line 58: should this say "expected format:" instead of "expected value:"  ?

sesdev/__init__.py
```diff
@@ -577,6 +581,23 @@ def _gen_settings_dict(version,
...
+                                        "synced folder paths",
+                                        (src, dst))
+            if not exists(src):
+                raise SettingValueError('synced_folder',
+                                        "Path to the source synced folder must exist",
+                                        src)
+
+        except ValueError:
+            raise SettingFormatError('synced_folder', "src:dst", folder)
+    settings_dict['synced_folder'] = [folder.split(':') for folder in synced_folder]
```
smithfarm, line 599: Though the options provided by the user typically enter the settings dictionary, at this point (here in `sesdev/__init__.py`) they are still options. Can we rename these new exceptions to `OptionValueError` and `OptionFormatError`, respectively (instead of `SettingValueError` and `SettingFormatError`)?

smithfarm: Allow me to make another attempt to explain why I prefer "option" instead of "setting" here. This code is validating user input, and the user is not inputting a "setting" - rather, the user's input is the value of an _option_ (in Click terminology).

seslib/__init__.py
```diff
@@ -461,6 +461,11 @@ def init_path_to_qa(cls, full_path_to_sesdev_executable):
         'help': 'Deploy SES using rook in CaasP',
         'default': False,
     },
+    'synced_folder': {
+        'type': list,
+        'help': 'Sync Folders to VM',
+        'default': list(),
```
smithfarm, line 467: Is there a particular reason why `list()` is better than `[]` here?

sesdev/__init__.py
```diff
@@ -577,6 +581,23 @@ def _gen_settings_dict(version,
     if not ceph_salt_deploy:
         settings_dict['ceph_salt_deploy'] = False
 
+    for folder in synced_folder:
+        try:
+            src, dst = folder.split(':')
+            if not all([isabs(x) for x in [src, dst]]):
+                raise SettingValueError('synced_folder',
```
smithfarm, line 588: Would the user find the error message easier to understand if we wrote `--synced-folder` here, instead of `synced_folder` (which IIRC is the name of an internal variable)?

sesdev/__init__.py
```diff
@@ -577,6 +581,23 @@ def _gen_settings_dict(version,
...
+                                        "Please provide absolute paths for "
+                                        "synced folder paths",
+                                        (src, dst))
+            if not exists(src):
+                raise SettingValueError('synced_folder',
```
smithfarm, line 593: Would the user find the error message easier to understand if we wrote `--synced-folder` here, instead of `synced_folder` (which IIRC is the name of an internal variable)?

seslib/__init__.py
jschmid1: I _personally_ like the lucidity of this syntax. An example:
While this is not so important for lists it's rather noticeable for dicts/sets.

``` python
In [1]: type([])                                                                
Out[1]: list

In [2]: type({})                                                                
Out[2]: dict

In [3]: type({1})                                                               
Out[3]: set
```

This is why I stick to explicit expressions. But in the end this is a matter of taste I guess. Your call.

smithfarm: The way I see it, either we (a) change `[]` to `list()` everywhere (many places) to conform with this, or (b) change `list()` to `[]` here.

I prefer (b), but I do see your point about using `dict()` instead of `{}`.

sesdev/__init__.py
```diff
@@ -577,6 +581,23 @@ def _gen_settings_dict(version,
...
+            src, dst = folder.split(':')
+            if not all([isabs(x) for x in [src, dst]]):
+                raise OptionValueError('--synced-folder',
+                                       "Please provide absolute paths for "
+                                       "synced folder paths",
+                                       (src, dst))
+            if not exists(src):
+                raise OptionValueError('--synced-folder',
+                                       "Path to the source synced folder must exist",
+                                       src)
```
smithfarm, line 595: Have you tested these error paths?

Instead of defining an exception in `seslib/exceptions.py`, it might be better to do something like this:

```
def _option_value_error(...):
    click.echo("error message")
    raise click.Abort()
```

smithfarm: Sorry, nevermind - it looks like raising an exception here will work. I should have tested it myself before making a comment.

seslib/exceptions.py
```diff
@@ -45,6 +45,20 @@ def __init__(self, setting, expected_type, value):
             .format(setting, expected_type, value, type(value)))
 
 
+class OptionValueError(SesDevException):
+    def __init__(self, option, message, value):
+        super(OptionValueError, self).__init__(
+            "Wrong value for option '{}'. {}. Actual value: '{}'"
```
smithfarm, line 51:
```
(venv) smithfarm@geras:~/sesdev> sesdev create octopus --single-node --stop-before-ceph-salt-config --synced-folder foo:bar
Wrong value for option '--synced-folder'. Please provide absolute paths for synced folder paths. Actual value: '('foo', 'bar')'
```

sesdev/__init__.py
```diff
@@ -577,6 +581,23 @@ def _gen_settings_dict(version,
...
         settings_dict['ceph_salt_deploy'] = False
 
+    for folder in synced_folder:
+        try:
+            src, dst = folder.split(':')
+            if not all([isabs(x) for x in [src, dst]]):
+                raise OptionValueError('--synced-folder',
+                                       "Please provide absolute paths for "
+                                       "synced folder paths",
+                                       (src, dst))
```
smithfarm, line 591:
```
(venv) smithfarm@geras:~/sesdev> sesdev create octopus --single-node --stop-before-ceph-salt-config --synced-folder foo:bar
Wrong value for option '--synced-folder'. Please provide absolute paths for synced folder paths. Actual value: '('foo', 'bar')'
```

Should it be `folder` here instead of `(src, dst)` ?

sesdev/__init__.py
```diff
@@ -577,6 +581,23 @@ def _gen_settings_dict(version,
...
+            if not exists(src):
+                raise OptionValueError('--synced-folder',
+                                       "Path to the source synced folder must exist",
+                                       src)
+
+        except ValueError:
+            raise OptionFormatError('--synced-folder', "src:dst", folder)
+    settings_dict['synced_folder'] = [folder.split(':') for folder in synced_folder]
```
smithfarm, line 599: hm, if the user gives multiple instances of `--synced-folder`, won't this cause earlier ones to be overwritten by later ones?

smithfarm: nevermind! I see the indentation now.

sesdev/__init__.py
jschmid1: sure, we can do that.

jschmid1:
> * `folder` instead of `(src, dst)` in the "Please provide absolute paths" error message

done

> * `[]` instead of `list()` (for uniformity)

done

smithfarm:
@jschmid1 When I merge this PR into the current master branch and then run

    sesdev create ses5 --single-node --non-interactive

I get the following error: https://gitlab.suse.de/snippets/468

My environment:

```
$ vagrant --version
Vagrant 2.2.7
$ cat /etc/os-release
NAME="SLES"
VERSION="15-SP1"
VERSION_ID="15.1"
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP1
```

jschmid1:
re-added `node.vm.synced_folder ".", "/vagrant", disabled: true`

Apparently this defaults to `disabled: false` and tries to nfs mount `.` to `/vagrant`.

smithfarm:
@jschmid1 Could you squash the two commits into one? Right now the first commit removes `synced_folder ".", "/vagrant", disabled: true` and the second one puts it back in...

In this case I think it makes sense to have the whole feature in a single commit.

jschmid1:
> @jschmid1 Could you squash the two commits into one? Right now the first commit removes `synced_folder ".", "/vagrant", disabled: true` and the second one puts it back in...
> 
> In this case I think it makes sense to have the whole feature in a single commit.

done

PR #249, opened 2020-04-14
README.md: add known pitfall "Failed to connect socket" error

smithfarm:


README.md
```diff
@@ -737,10 +742,35 @@ done
...
+
+```
+libvirt.libvirtError: Failed to connect socket to
+'/var/run/libvirt/libvirt-sock': No such file or directory
+```
+
+#### Analysis
+
+When told to use remote libvirt/SSH, sesdev expects that there won't be any
+libvirtd instance running locally. This Python traceback is displayed when
```
jschmid1, line 760: lgtm, but this really sucks :/

smithfarm: what sucks? The fact that you can't connect to remote libvirtd when libvirtd is running locally, or the fact that sesdev barfs out a Traceback instead of displaying a more elegant error message?

jschmid1: > The fact that you can't connect to remote libvirtd when libvirtd is running locally,

this

toabctl: Is that a limitation sesdev? Or vagrant-libvirt? Because `virt-manager` can handle that case easily.

smithfarm: I don't know.

Issue #252, opened 2020-04-15
A command that returns the number of deployments


smithfarm:
We have `sesdev list`, but for testing purposes it would be nice to have a way to "programatically" (i.e. from a bash script like `contrib/standalone.sh`) get the number of existing deployments in the list.

This could be done either by

* implementing a `--format json` option to `sesdev list` (then the user could do something like `sesdev list --format json | jq -r ". | length"`)

or

* implementing a `--number` option which would cause `sesdev list` to print, e.g., "3" if there are three deployments

PR #259, opened 2020-04-15
sesdev: implement "sesdev list --format json"

smithfarm:
Fixes: https://github.com/SUSE/sesdev/issues/252

---

contrib/standalone.sh: do not start testing in unclean env 

---

sesdev: use prettytable to generate "sesdev list" table 

---

### DIFF
--- a/contrib/standalone.sh
+++ b/contrib/standalone.sh
@@ -117,6 +117,12 @@ if [ "$ALL" ] ; then
     SES7="--ses7"
 fi
 
+if [ "$(sesdev list --format json | jq -r '. | length')" != "0" ] ; then
+    echo "ERROR: detected existing deployments"
+    echo "(This script expects a clean environment -- i.e., \"sesdev list\" must be empty)"
+    exit 1
+fi
+
 if [ "$SES5" ] ; then
     run_cmd sesdev box remove --non-interactive sles-12-sp3 || true
     # deploy ses5 without igw, so as not to hit https://github.com/SUSE/sesdev/issues/239
--- a/sesdev.spec
+++ b/sesdev.spec
@@ -36,11 +36,13 @@ BuildRequires:  python3-setuptools
 %if 0%{?suse_version}
 Requires:       python3-Jinja2 >= 2.10.1
 Requires:       python3-libvirt-python >= 5.1.0
+Requires:       python3-PrettyTable
 Requires:       python3-PyYAML >= 3.13
 %endif
 %if 0%{?el8} || 0%{?fedora}
 Requires:       python3-jinja2 >= 2.10.1
 Requires:       python3-libvirt >= 5.1.0
+Requires:       python3-prettytable
 Requires:       python3-pyyaml >= 3.13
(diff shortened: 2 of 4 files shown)


smithfarm:

toabctl:
> This passed a `contrib/standalone.sh --full` run.

Should we just use that wrapper script in the Jenkinsfile then? Or is there any major diff between the script and the code we have in Jenkinsfile.integration ?

smithfarm:
> Should we just use that wrapper script in the Jenkinsfile then? Or is there any major diff between the script and the code we have in Jenkinsfile.integration ?

`contrib/standalone.sh --full` takes 4-6 hours to run, so probably not appropriate for Jenkins.

sesdev.spec
```diff
@@ -44,6 +44,7 @@ Requires:       python3-libvirt >= 5.1.0
 Requires:       python3-pyyaml >= 3.13
 %endif
 Requires:       python3-click >= 6.7
+Requires:       python3-prettytable
```
toabctl, line 47: At least on openSUSE , this is `python3-PrettyTable`

contrib/standalone.sh
```diff
@@ -117,6 +117,12 @@ if [ "$ALL" ] ; then
     SES7="--ses7"
 fi
 
+if [ "$(sesdev list --format json | jq -r '. | length')" != "0" ] ; then
```
toabctl, line 120: Isn't `sesdev create` itself checking that a deployment with a similar name doesn't exist?

sesdev.spec
toabctl: Also, the last PrettyTable release is from 2013. Maybe use `tabulate` (https://pypi.org/project/tabulate/) ?

setup.cfg
```diff
@@ -25,6 +25,7 @@ install_requires =
     Click >= 6.7
     Jinja2 >= 2.10.1
     libvirt-python >= 5.1.0
+    prettytable
```
toabctl, line 28: PrettyTable (see pypi name)

contrib/standalone.sh
smithfarm: yes, `sesdev create` checks that and when it finds the existing deployment it returns an error and that stops the `contrib/standalone.sh` run, which is undesirable.

setup.cfg
smithfarm: apparently, `prettytable` is an alias, because this works fine?

sesdev.spec
smithfarm: Though it was released in 2013, it works great.

smithfarm:
By the way, Ceph itself uses PrettyTable.

setup.cfg
toabctl: On `pypi`, names are case-insensitive. so that just works. But you also have `Click`, not `click`. So using the "correct" name would be nice

smithfarm: Thanks, @toabctl - TIL!

PR #262, opened 2020-04-17
qa: fix maybe_wait_for_mdss test

smithfarm:
Since cephadm/orchestrator does Day 1 operations asynchronously, when an
Octopus/Pacific cluster is coming up it can take some time for the MDS
metadata to be updated. And, in fact, there is no guarantee that the MDS
metadata will be present before "ceph orch ls --service-type mds" starts
reporting the MDS as "running".

Also, do some general clean up of the qa bash scripting, including further
work to make it ShellCheck-clean and facilitate running ShellCheck on the
code.


smithfarm:

PR #267, opened 2020-04-18
Rename --deepsea-cli and --ceph-salt-deploy options to --salt

smithfarm:
With this patch, when we want "salt ... state.apply" (SES7) or "salt-run
state.orch ..." (SES6, SES5), we simply give the option --salt. This is
much more straightforward (and easier to remember) than the previous
method.

Note: this patch does *not* change the default behavior, which is to use
"ceph-salt deploy --non-interactive" (SES7) or "deepsea stage run ..."
(SES6, SES5). The patch only renames the options used to trigger the
non-default behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ the VMs and run the deployment scripts.
 * [Usage](#usage)
    * [Create/deploy a Ceph cluster](#createdeploy-a-ceph-cluster)
       * [On a remote libvirt server via SSH](#on-a-remote-libvirt-server-via-ssh)
+      * [Using salt instead of DeepSea/ceph-salt CLI](#using-salt-instead-of-deepseaceph-salt-cli)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -358,6 +359,15 @@ libvirt_host: <hostname|ip address>
 Note that passwordless SSH access to this user@host combination needs to be
 configured and enabled.
 
+#### Using salt instead of DeepSea/ceph-salt CLI
+
+By default, sesdev will use the DeepSea CLI to run the DeepSea Stages (ses5,
+nautilus, ses6) or the "ceph-salt" command to apply the ceph-salt Salt Formula
+(ses7, octopus).
```
ricardoasmarques, line 366: Also for pacific - https://github.com/SUSE/sesdev/blob/master/seslib/__init__.py#L213

smithfarm: done

smithfarm:

Issue #266, opened 2020-04-18
click 6.7: "TypeError: FOO got an unexpected keyword argument 'force'"


smithfarm:
Latest `sesdev` installed from RPM (not reproducible on master branch installed from source):

```
$ sesdev destroy -f '*'
Traceback (most recent call last):
  File "/usr/bin/sesdev", line 11, in <module>
    load_entry_point('sesdev==1.1.10+1587124098.ga2f8189', 'console_scripts', 'sesdev')()
  File "/usr/lib/python3.6/site-packages/sesdev/__init__.py", line 20, in sesdev_main
    cli(prog_name='sesdev')
  File "/usr/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
TypeError: destroy() got an unexpected keyword argument 'force'
```

```
$ sesdev --version
sesdev 1.1.10+1587124098.ga2f8189
```

---

```
+ sesdev create ses5 --non-interactive --roles '[master,storage,mon,mgr,mds,rgw,nfs]' --qa-test ses5-1node
Traceback (most recent call last):
  File "/usr/bin/sesdev", line 11, in <module>
    load_entry_point('sesdev==1.1.10+1587124098.ga2f8189', 'console_scripts', 'sesdev')()
  File "/usr/lib/python3.6/site-packages/sesdev/__init__.py", line 20, in sesdev_main
    cli(prog_name='sesdev')
  File "/usr/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
[...]

PR #268, opened 2020-04-20
sesdev: support Click 6.7

smithfarm:
openSUSE Leap 15.1 and SLE-15-SP1 are still on Click 6.7, so we have to
continue supporting it.

Fixes: https://github.com/SUSE/sesdev/issues/266


smithfarm:

smithfarm:
This one is urgent: it's only a matter of time before someone notices that the sesdev RPM is broken :-/

Issue #269, opened 2020-04-20
dashboard tunnel fails


jschmid1:
with:

`sesdev create ses7 --stop-before-ceph-salt-config`

```
➜  ~ sesdev tunnel ses7 dashboard
Opening tunnel to service 'dashboard' on node 'master'...
Node '' does not exist in this deployment
➜  ~ sesdev --version            
sesdev 1.1.10
```

jschmid1:
this https://github.com/SUSE/sesdev/blob/c2b5a851f4007d61d92d7e470ff8e74122dc4d0f/seslib/__init__.py#L1725 
returns `''`, hence isn't caught by the except block.

jschmid1:
When nodes/roles are assigned:

```
master:~ # ceph-salt config ceph_cluster/roles  ls
o- roles ...................................................................................................... [...]
  o- admin ............................................................................................. [Minions: 1]
  | o- master.ses7.com ..................................................................... [Other roles: bootstrap]
  o- bootstrap .................................................................................... [master.ses7.com]
```

I see that the `master` node does not have the mentioned roles assigned.

```
(Pdb) _node.name
'master'
(Pdb) _node.roles
['master', 'client', 'prometheus', 'grafana']
```

while `node1`, which doesn't have the roles assigned (in `config`) shows _does_ report it in the `Node` object.

```
(Pdb) _node.name
'node1'
(Pdb) _node.roles
['bootstrap', 'storage', 'mon', 'mgr', 'rgw', 'igw']
```

This leads to wrong targeting in the next block.

https://github.com/SUSE/sesdev/blob/c2b5a851f4007d61d92d7e470ff8e74122dc4d0f/seslib/__init__.py#L1720-L1722

and to a consecutive failure since the `ceph` binary is not installed on this node yet.

smithfarm:
This is an interesting bug report. Thanks, @jschmid1, for opening it.

I understand *what* is happening, but I'm not sure how to fix it. Here is the "what":

When you don't specify any roles, `sesdev` gives you the default roles, which put `bootstrap` on node1.

You, however, manually told `ceph-salt` that "master" is the bootstrap node.

Another problem is that `sesdev tunnel ses7 dashboard` is not savvy. If you don't provide the name of the node, you get `master` by default.

You can work around this issue by:

1. knowing which node "cephadm bootstrap" was run on
2. providing that node on the `sesdev tunnel dashboard NODE` command line

smithfarm:
[...]

PR #274, opened 2020-04-21
seslib: more robust dashboard tunnelling

smithfarm:
We assume that "master" will be an "admin" node, and hence that
"ceph mgr services" will work there.

This commit changes the logic so the output of "ceph mgr services" is
parsed in the Python code, instead of by piping the output to "jq" and
"sed" on the cluster node like we were doing before.

It also adds a regression test for the dashboard tunnelling feature,
which also asserts that at least *some* form of dashboard downstream
branding has been applied in SES5, SES6 and SES7.

Also take the opportunity to make "contrib/standalone.sh"
ShellCheck-clean.

Fixes: https://github.com/SUSE/sesdev/issues/269

### DIFF
--- a/contrib/standalone.sh
+++ b/contrib/standalone.sh
@@ -22,13 +22,15 @@
 
 set -x
 
-SCRIPTNAME=$(basename ${0})
+SCRIPTNAME="$(basename "${0}")"
 FINAL_REPORT="$(mktemp)"
+TEMP_FILE="$(mktemp)"
 
 function final_report {
-    echo -en "\n=====================================================================\n" >> $FINAL_REPORT
-    cat $FINAL_REPORT
-    rm $FINAL_REPORT
+    echo -en "\n=====================================================================\n" >> "$FINAL_REPORT"
+    cat "$FINAL_REPORT"
+    rm "$FINAL_REPORT"
+    rm "$TEMP_FILE"
     exit 0
 }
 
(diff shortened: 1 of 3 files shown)


smithfarm:
Interestingly enough, `sesdev tunnel DEP_ID dashboard` works on a default `ses5` deployment:

```
+ test_tunnel_dashboard ses5-4node
+ local dep_id=ses5-4node
+ sleep 30
+ timeout 40s sesdev tunnel ses5-4node dashboard
Opening tunnel to service 'dashboard' in deployment 'ses5-4node'...
Service 'dashboard' was not found in this deployment
+ curl --silent --insecure https://127.0.0.1:8443
+ grep 'SUSE Enterprise Storage'
  <title>SUSE Enterprise Storage</title>
+ sleep 15
```

contrib/standalone.sh
```diff
@@ -139,7 +155,9 @@ if [ "$SES5" ] ; then
     run_cmd sesdev destroy --non-interactive ses5-1node
     run_cmd sesdev create ses5 --non-interactive --roles [master,client,openattic],[storage,mon,mgr,rgw],[storage,mon,mgr,mds,nfs],[storage,mon,mgr,mds,rgw,nfs] ses5-4node
     run_cmd sesdev qa-test ses5-4node
+    test_tunnel ses5-4node openattic http 8080 Redirecting
     run_cmd sesdev destroy --non-interactive ses5-4node
+    curl --silent --insecure "http://127.0.0.1:8080" || true
```
toabctl, line 160: Why is this needed?

smithfarm: I would like to assert that tunnel is gone after `test_tunnel` completes, but this line was just a temporary experiment. It is reworked in the commit I will be pushing presently.

contrib/standalone.sh
```diff
@@ -63,25 +65,54 @@ function usage {
...
+    [ -z "$port" ]          && port="8443"
+    [ -z "$search_string" ] && search_string="SUSE Enterprise Storage"
+    timeout 40s sesdev tunnel "$dep_id" "$service" &
+    sleep 30
+    curl --silent --insecure "$protocol://127.0.0.1:$port" | tee "$TEMP_FILE"
+    grep "$search_string" "$TEMP_FILE"
+    sleep 15
+}
+
+function tunnel_gone {
```
toabctl, line 102: I don't understand why we want to test this. It's testing that the `timeout` command is working and kills the process correctly. Do we want to test `timeout` here? Or if we want to test that `sesdev tunnel` is reacting to the signal correctly, can we use `kill %1` instead of `timeout` and `sleep`? That would safe us at least 15 seconds per run.
I'm also fine with just merging. Just thinking loud...

contrib/standalone.sh
```diff
@@ -157,7 +190,9 @@ if [ "$SES6" ] ; then
     run_cmd sesdev destroy --non-interactive ses6-1node
     run_cmd sesdev create ses6 --non-interactive ses6-4node
     run_cmd sesdev qa-test ses6-4node
+    test_tunnel ses6-4node
     run_cmd sesdev destroy --non-interactive ses6-4node
+    curl --silent --insecure "https://127.0.0.1:8443" || true
```
toabctl, line 195: shouldn't that be `tunnel_gone ...`?

contrib/standalone.sh
smithfarm: Hi Tom, hopefully https://github.com/SUSE/sesdev/issues/276 answers the question.

contrib/standalone.sh
smithfarm: good catch! thanks

PR #275, opened 2020-04-21
deepsea: "openattic" role needs Stage 4

smithfarm:
Before this fix, we might skip Stage 4 even though the "openattic" role
is given. In such a case, openATTIC would not get deployed, which is
a bug.


smithfarm:
Reproducer:

    sesdev create ses5 --roles="[master,storage,mon,mgr,openattic]"

Issue #278, opened 2020-04-22
provision.sh fails transiently with "scp: /tmp/ready: No such file or directory"


smithfarm:
I saw a 4-node ses7 deployment fail in this code:

```
while : ; do
  PROVISIONED_NODES=`ls -l /tmp/ready-* 2>/dev/null | wc -l`
  echo "waiting for nodes (${PROVISIONED_NODES}/{{ nodes|length }})";
  [[ "${PROVISIONED_NODES}" != "{{ nodes|length }}" ]] || break
  sleep 2;
{% for node in nodes %}
  scp -o StrictHostKeyChecking=no {{ node.name }}:/tmp/ready /tmp/ready-{{ node.name }};
{% endfor %}
done
```

The relevant section of the console log is:

```
    master: ++ scp -o StrictHostKeyChecking=no node1:/tmp/ready /tmp/ready-node1
    master: Warning: Permanently added 'node1,10.20.60.201' (ECDSA) to the list of known hosts.
    node3: .
    master: ++ scp -o StrictHostKeyChecking=no node2:/tmp/ready /tmp/ready-node2
    node3: .
    master: Warning: Permanently added 'node2,10.20.60.202' (ECDSA) to the list of known hosts.
    master: ++ scp -o StrictHostKeyChecking=no node3:/tmp/ready /tmp/ready-node3
    node3: .
    master: Warning: Permanently added 'node3,10.20.60.203' (ECDSA) to the list of known hosts.
    node3: .
    master: scp: /tmp/ready: No such file or directory
```

Further examination of the log yielded the following information:

* `++ touch /tmp/ready` is present for all nodes except `node3`

My theory is there is some kind of race condition in the `while` loop in the code posted above.

smithfarm:
Ah, now I see this is a regression caused by:

c0d5df13b20635ad8a7a0075f74b21800cd10ffc
provision.sh: abort if any command fails

PR #279, opened 2020-04-22
provision.sh: refactor PROVISIONED_NODES while loop

smithfarm:
If all the "/tmp/ready-{{ node.name }}" files don't get created in the
first iteration of the loop, the scp fails and that causes the whole
script to fail because it is running under "set -e".

(This while loop was written before we had "set -e" at the top of the
script.)

Since we're touching the code of the loop anyway, take this opportunity
to refactor it.

Fixes: c0d5df13b20635ad8a7a0075f74b21800cd10ffc
Fixes: https://github.com/SUSE/sesdev/issues/278


smithfarm:

smithfarm:
Triggered RPM build in https://build.opensuse.org/package/show/filesystems:ceph/sesdev and will run `contrib/standalone.sh --full` on that.

smithfarm:
Tested manually.

PR #285, opened 2020-04-27
ceph_salt_deployment: really sync clocks

smithfarm:
Before, we had some very simplistic code for syncing the clocks with the
time server before deploying additional MONs. This code was not actually
syncing the clocks.

### DIFF
--- a/seslib/templates/ceph-salt/ceph_salt_deployment.sh.j2
+++ b/seslib/templates/ceph-salt/ceph_salt_deployment.sh.j2
@@ -70,7 +70,8 @@ ceph-salt config /ssh/ generate
 ceph-salt config /containers/images/ceph set {{ image_path }}
 {% endif %}
 ceph-salt config /time_server/server_hostname set {{ master.fqdn }}
-ceph-salt config /time_server/external_servers add 0.pt.pool.ntp.org
+{% set external_timeserver = "pool.ntp.org" %}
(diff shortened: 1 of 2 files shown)


seslib/templates/ceph-salt/ceph_salt_deployment.sh.j2
```diff
@@ -103,12 +104,55 @@ stdbuf -o0 ceph-salt -ldebug apply --non-interactive
 {% if ceph_salt_deploy_mons %}
 {% if mon_nodes > 1 %}
 TIME_SERVER="$(ceph-salt export | jq -r .time_server.server_host)"
-salt "$TIME_SERVER" cmd.run 'bash -c "chronyc makestep && chronyc waitsync"'
-{% for node in nodes %}
-if [ "{{ node.fqdn }}" != "$TIME_SERVER" ] ; then
-    salt "{{ node.fqdn }}" cmd.run 'bash -c "chronyc makestep && chronyc waitsync"'
+{% set chrony_script = "chrony_sync_start.sh" %}
```
toabctl, line 107: Maybe move that into an extra template and include it? eg. `seslib/templates/ceph-salt/ceph_salt_chrony.sh.j2` ?

smithfarm: This code is going to be deleted soon (as soon as I can implement it in ceph-salt), so I would say not worth the trouble?

Issue #299, opened 2020-05-04
CaaSP clusters are not actually provisioned


colder-is-better:
Currently, whenever we use `caasp4` as a target for the `create` command, we indeed end up with a set of SLE15SP1 nodes but with no _actual_ CaaSP 4 cluster. The nodes are prepared and ready for CaaSP 4 (i.e., the product, the patterns, and all required repositories/modules are there), but the CaaSP 4 cluster itself is not provisioned. Manual provisioning of CaaSP 4 is possible and works fine, but it would be nicer to issue a command like

```
sesdev create caasp4 --domain local.is \
	--num-disks 2 --disk-size 64 --ram 6 --cpus 4 \
	--roles="[master], [worker, storage], [worker, storage], [worker, storage]" \
	k8sbaby --non-interactive
```

and after a while have a fully working CaaSP 4 cluster. Additionally, including `--deploy-ses` on a command like the above would automatically install Rook over CaaSP 4. (Command `--deploy-ses` is already there, but for now it looks like it has no effect.)

toabctl:
cc @capri1989 iirc you did some implementation during hackweek, right?

capri1989:
I did the implementation during hackweek, yeah and I would need to look at it again. Will put on my backlog but can't promise when this will be done.

capri1989:
https://github.com/SUSE/sesdev/pull/302

smithfarm:
@colder-is-better Can this be closed now?

colder-is-better:
It sure can -- thank you!

---
I survive (mostly) by tapping keys
Always remember: Colder Is Better (TM)

PR #302, opened 2020-05-06
caasp master provision fix

capri1989:
This is a fix for https://github.com/SUSE/sesdev/issues/299

As @smithfarm removed the multi-master deployment functionality from sesdev (caasp part) this is a quick fix to get the caasp deployment up and running again. If we ever need the possibility to do multi-master with caasp we could add it back. 

### DIFF
--- a/seslib/templates/caasp/master.sh.j2
+++ b/seslib/templates/caasp/master.sh.j2
@@ -1,11 +1,11 @@
 
 zypper --non-interactive install --type pattern SUSE-CaaSP-Management
 
-{% if node.name == 'master1' %}
+{% if node.name == 'master' %}
 
-function wait_for_masters_ready {
-    printf "Waiting for masters to be ready"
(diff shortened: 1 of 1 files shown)


seslib/templates/caasp/master.sh.j2
```diff
@@ -1,11 +1,11 @@
 
 zypper --non-interactive install --type pattern SUSE-CaaSP-Management
 
-{% if node.name == 'master1' %}
+{% if node.name == 'master' %}
 
-function wait_for_masters_ready {
-    printf "Waiting for masters to be ready"
-    until [[ $(kubectl get nodes 2>/dev/null | egrep -c "master[0-9]\s+Ready") -eq {{node_manager.get_by_role('master') | length}} ]]; do
+function wait_for_master_ready {
```
smithfarm, line 6: do you also need to change the invocation of this function on line 70?

capri1989: Good catch, thx. Fixed

PR #303, opened 2020-05-06
seslib: drop Containers module from SES7 deployment

smithfarm:


smithfarm:
The Jenkins failure here is interesting. This behavior does not reproduce outside of OpenStack.

We have known all along that, from time to time, Ceph behaves in odd ways when run in an OpenStack environment, but it's very difficult to put a handle on, since the odd behavior typically goes away after a time. I suspect that will be the case here, as well: something will change over at OVH, and the idiopathic behavior will just "go away".

PR #311, opened 2020-05-13
provision.sh: remove curl RPM from the environment

smithfarm:
Like "which", "curl" is an RPM that is often silently assumed to be
present, leading to bugs like:

https://tracker.ceph.com/issues/44498


smithfarm:
So far, this has revealed an undeclared curl dependency in SES5 DeepSea Stage 2 (`ceph.monitoring`)

PR #24, opened 2019-12-02
Implement replace-mgr-modules subcommand

tspmelo:
Signed-off-by: Tiago Melo <tmelo@suse.com>

---

TO DO:

- [x] add README section explaining how to use it

PR #331, opened 2020-06-04
Allow user to re-install ceph-salt from source

ricardoasmarques:
This PR adds the `replace-ceph-salt` command that allows the user to re-install ceph-salt from a given source (by using `rsync` to update files):

```
# sesdev replace-ceph-salt octopus --local ~/projects/ceph-salt
```
This PR was inspired by https://github.com/SUSE/sesdev/pull/24, so all different options to fetch code should be consistent between both `replace-ceph-salt` and `replace-mgr-modules` commands.

For now, only `local` code is supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ The Jenkins CI tests that `sesdev` can be used to deploy a single-node Ceph
       * [Run "make check" on openSUSE Leap 15.2 from upstream "octopus" branch](#run-make-check-on-opensuse-leap-152-from-upstream-octopus-branch)
       * [Run "make check" on SLE-15-SP2 from downstream "ses7" branch](#run-make-check-on-sle-15-sp2-from-downstream-ses7-branch)
       * [Other "make check" scenarios](#other-make-check-scenarios)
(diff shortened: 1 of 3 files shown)


smithfarm:
It looks like it will replace ceph-salt regardless of whether it was installed from source or from RPM - nice!

README.md
```diff
@@ -600,6 +601,14 @@ $ sesdev tunnel <deployment_id> dashboard
 
 The command will output the URL that you can use to access the dashboard.
 
+### Replace ceph-salt
+
+It's possible to replace ceph-salt from source.
+
+```
+$ sesdev replace-ceph-salt --local <path>
+```
```
smithfarm, line 610: Would the following be better?

For deployments that used ceph-salt, it's possible to replace the ceph-salt installed by sesdev with a different one:

    $ sesdev replace-ceph-salt --local <path> <deployment_id>

Assuming <path> points to ceph-salt source code, the command will work regardless of whether ceph-salt was originally installed from source or from RPM.

ricardoasmarques: Done

PR #333, opened 2020-06-06
seslib: move boilerplate ssh options into a staticmethod

smithfarm:
At the beginning, Deployment had just one ssh method so there was no
code duplication.

Then, the scp feature was added and the ssh boilerplate options got
duplicated.

Recently, the rsync feature was added, which uses the same ssh
boilerplate options. At that point, the same exact set of boilerplate
options appeared in three different locations.

If we ever wanted to modify the boilerplate options, we would have to
modify it in at least three places. To avoid that, move the boilerplate
options into a dedicated, static method.

### DIFF
--- a/seslib/__init__.py
+++ b/seslib/__init__.py
@@ -1601,13 +1601,33 @@ class Deployment():
 
         return (host_is_source, host_is_destination, name, source_path, destination_path)
 
+    @staticmethod
+    def __boilerplate_ssh_options(string=False):
+        retval = None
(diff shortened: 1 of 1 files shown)


smithfarm:
@tspmelo I will test that this doesn't break ssh or scp. Could you test that it doesn't break rsync?

tspmelo:
tested this PR with `replace-mgr-modules` and `rsync` and it worked.

PR #345, opened 2020-06-16
cleanup: nodes_with_role

smithfarm:
We have some messy code in ceph_salt_deployment.sh which populates some shell variables with the names of nodes that are supposed to have a given role.

This changeset moves this code into `seslib/__init__.py` by populating python arrays with the same information and converting these into comma-separated strings which are sent into ceph_salt_deployment.sh via the Jinja context.

This also lays some groundwork for further PRs expanding the QA tests so they not only vet the expected number of daemons of various types, but also assert that daemons are actually running on the expected nodes.

### DIFF
--- a/qa/common/helper.sh
+++ b/qa/common/helper.sh
@@ -21,7 +21,23 @@ function _copy_file_from_minion_to_master {
 
 function _first_x_node {
     local ROLE=$1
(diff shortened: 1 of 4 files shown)


smithfarm:
This was included in a `contrib/standalone.sh` run, which validated that it is fully functional (all octopus/ses7/pacific tests passed, for example), though the run itself contained some failures that were fully attributable to a different PR in the integration branch.

PR #346, opened 2020-06-16
qa/common: make "ceph orch ps" test more resilient

smithfarm:
Turns out it may take some time before the "ceph orch ps" output shows
all the daemons that are expected to be running in the cluster. Add
a try-wait loop to this test.

### DIFF
--- a/qa/common/common.sh
+++ b/qa/common/common.sh
@@ -564,74 +564,99 @@ function number_of_services_expected_vs_orch_ls_test {
     fi
 }
 
+function _orch_ps_test {
+    local success
+    success="yes"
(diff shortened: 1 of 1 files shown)


smithfarm:
This one is weird. In two consecutive runs I saw this "orch ps" test report insufficient numbers of MONs and MGRs. One time it showed 0 instead of 3, and another time it showed 1 instead of 3. But both times when I SSHed in and ran "orch ps", it showed 3 MONs and 3 MGRs as expected.

Since then, though, I have run the test quite a few times and have not seen the behavior repeated. 

I guess it makes sense to merge it, just to lower the probability of false failure.

PR #363, opened 2020-06-22
ceph_salt_deployment.sh: reduce number of "ceph orch apply" calls

smithfarm:
Instead of calling "ceph orch apply <service>" for each type of daemon
we want to deploy,

1. make a big service spec yaml blob for {mon,mgr,osd,mds} and run
   "ceph orch apply" on it
2. wait for all the OSDs to join the cluster
3. make a second big service spec yaml blob for {rgw,nfs} (and others to
   be added later) and run "ceph orch apply" on it

The rationale for this is:

1. this is more idiomatic for cephadm, and
2. it brings us closer to the goal of generating one, big blob of YAML for the whole cluster to be passed to "cephadm bootstrap" via "ceph-salt config/apply".

### DIFF
--- a/contrib/standalone.sh
+++ b/contrib/standalone.sh
@@ -48,8 +48,8 @@ function usage {
     echo
     echo "Options:"
     echo "    --help                   Display this usage message"
-    echo "    --ceph-salt-from-source  Install ceph-salt from source"
-    echo "                             (default: from package)"
+    echo "    --ceph-salt-branch       Branch from which to install ceph-salt from source"
+    echo "    --ceph-salt-repo         Repo from which to install ceph-salt from source"
     echo "    --full                   Run all tests, including makecheck"
     echo "    --makecheck              Run makecheck (install-deps.sh, actually)"
     echo "                             tests"
@@ -120,14 +120,16 @@ function tunnel_gone {
 }
 
 TEMP=$(getopt -o h \
---long "help,caasp4,ceph-salt-from-source,full,makecheck,nautilus,no-stop-on-failure,octopus,pacific,ses5,ses6,ses7" \
+--long "help,caasp4,ceph-salt-branch:,ceph-salt-repo:,full,makecheck,nautilus,no-stop-on-failure,octopus,pacific,ses5,ses6,ses7" \
 -n 'standalone.sh' -- "$@") || ( echo "Terminating..." >&2 ; exit 1 )
 eval set -- "$TEMP"
 
 # process command-line options
 NORMAL_OPERATION="not_empty"
 CAASP4=""
-CEPH_SALT_FROM_SOURCE=""
+CEPH_SALT_OPTIONS=()
+CEPH_SALT_BRANCH=""
+CEPH_SALT_REPO=""
 FULL=""
 MAKECHECK=""
 NAUTILUS=""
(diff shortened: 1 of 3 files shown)


smithfarm:
So, "single service spec" deployment (i.e., deployment of a whole cluster with a single invocation of `ceph orch apply`) is not possible because RGW deployment requires that we run the following commands **before** `ceph orch apply`:

```
radosgw-admin realm create --rgw-realm=default --default
radosgw-admin zonegroup create --rgw-zonegroup=default --master --default
radosgw-admin zone create --rgw-zonegroup=default --rgw-zone=default --master --     └ default
```

... and the first command hangs because it needs a running cluster (with OSDs) to complete.

So, we need at least two invocations of `ceph orch apply`: one for the OSDs, and a second for the rest of the cluster.

jschmid1:
That is also true for NFS which needs an active pool.

What about splitting the spec in two.

1) Core & Independent services

In between you could do the necessary invocations and then continue

2) Gateway & Dependent services

sebastian-philipp:
> That is also true for NFS which needs an active pool.
> 
> What about splitting the spec in two.
> 
> 1. Core & Independent services
> 
> In between you could do the necessary invocations and then continue
> 
> 1. Gateway & Dependent services

Right. 

At least for RGW, we should add those commands to mgr/cephadm in any case. Cause the current behavior is just bad UX.

for NFS, this needs some more discussion with the NFS folks. I mean others will run into the same problem that it's impossible to have a big day1 cluster.yaml including cephfs and NFS.

smithfarm:
> I mean others will run into the same problem that it's impossible to have a big day1 cluster.yaml including cephfs and NFS.

Not only is it impossible, but since it just hangsit's not clear what the problem

smithfarm:
After running the first `ceph orch apply`, we will have to either

1. insert a sleep (meh), or
2. insert a try-wait loop waiting for all the OSDs to come up

before attempting to create the RGW, otherwise the `radosgw-admin` command will hang.

sebastian-philipp:
https://tracker.ceph.com/issues/43681

smithfarm:
Finally got it working.

smithfarm:
All the 4-node deployments are showing the following test failure:

```
    master: WWWW: number_of_services_expected_vs_orch_ls_test
    master: MGR services (orch ls/expected): null
    master: null/3
    master: MON services (orch ls/expected): null
    master: null/3
    master: MDS services (orch ls/expected): 2/2
    master: RGW services (orch ls/expected): 2/2
    master: NFS services (orch ls/expected): 2/2
    master: ERROR: Detected disparity between expected number of services and "ceph orch ls"!
    master: WWWW: number_of_services_expected_vs_orch_ls_test: FAIL
```

Needs further investigation.

smithfarm:
Either there is a bug in cephadm, or I am setting the MON and MGR service_id incorrectly:

```
master:~ # ceph orch ls --service-type mgr
NAME                                    RUNNING  REFRESHED  AGE  PLACEMENT          IMAGE NAME                                                                       IMAGE ID      
mgr.sesdev_mgr_deployment                   0/3  -          -    node1,node2,node3  <unknown>                                                                        <unknown>     
mgr.sesdev_mgr_deployment.node2.usqbyy      4/0  6m ago     -    <unmanaged>        registry.opensuse.org/filesystems/ceph/octopus/upstream/images/ceph/ceph:latest  a2c23ec01b1e  
```

(I'm leaning towards thinking this is a bug, because the OSD service does the same thing and that was recognized as a bug.)

Issue #304, opened 2020-05-07
Explicit "--roles" specification ignored when "--single-node" option given


smithfarm:
The following behavior is expected, but might be surprising:

```
(venv) smithfarm@vanguard1:~/sesdev> sesdev create ses6 --single-node --roles "[master,storage,mon,mgr]"
=== Creating deployment "ses6_mini" with the following configuration ===
1 Deployment VMs:
  -- master:
     - OS:               sles-15-sp1
     - ses_version:      ses6
     - deployment_tool:  deepsea
     - roles:            ['grafana', 'igw', 'master', 'mds', 'mgr', 'mon', 'nfs', 'prometheus', 'rgw', 'storage']
```

i.e.: the `--roles` specification on the command line was ignored.

We should error-exit when `--single-node` and `--roles` are both present, instead of silently ignoring `--roles`.

PR #376, opened 2020-07-02
sesdev: raise exception if --roles combined with --single-node

smithfarm:
Since --single-node silently clobbers --roles, it's not a good
idea to let the user set them both at the same time.

Fixes: https://github.com/SUSE/sesdev/issues/304


smithfarm:
```
$ sesdev create octopus --roles "[master,mon,mgr,storage]" --single-node
The --roles and --single-node options are mutually exclusive. One may be given, or the other, but not both at the same time.
```

PR #378, opened 2020-07-02
Update the example Deployment id in sesdev help

supriti:
Valid characters for hostnames are ASCII(7) letters from a to z, the digits from 0 to 9, and the hyphen (-).

Signed-off-by: supriti <supriti.singh@suse.com>


smithfarm:
Thanks for the fix, @supriti ! 

Nowadays the deployment ID is actually optional ... you can also do, e.g., `sesdev create ses6` - in this example, the deployment ID will default to "ses6".

PR #382, opened 2020-07-10
sesdev: use Luminous roles for SES5

bk201:
Fix the error `The DeepSea version used in SES5 does not recognize 'prometheus' or 'grafana' as roles in policy.cfg.` when creating a cluster with default roles.

Signed-off-by: Kiefer Chang <kiefer.chang@suse.com>

PR #384, opened 2020-07-11
sesdev: implement --dry-run for create commands

smithfarm:
"sesdev create" with the `--dry-run` option implemented by this commit
will bail out before creating any VMs.

In practice, this means immediately prior to triggering `Deployment.start()`

Useful for testing.


smithfarm:
@bk201 This should prevent regressions like the one you just fixed in #382

PR #387, opened 2020-07-14
'sesdev replace-mgr-modules' should also replace 'bin/cephadm'

ricardoasmarques:
In addition to replace MGR modules, 'bin/cephadm' should also be replaced in order to test cephadm patches / PRs.

### DIFF
--- a/seslib/deployment.py
+++ b/seslib/deployment.py
@@ -1129,9 +1129,11 @@ class Deployment():
             print("Fetching branch '{}' from '{}'...".format(branch, repo))
 
         master_node = 'master'
+        all_nodes = []
         mgr_nodes = []
 
         for _node in self.nodes.values():
+            all_nodes.append(_node.name)
             if _node.has_role('mgr'):
                 mgr_nodes.append(_node.name)
 
@@ -1148,6 +1150,7 @@ class Deployment():
             tools.run_sync(ssh_cmd)
 
         print("Fetching...")
+        # Fetch mgr modules
         if local:
             local_path = "{}/src/pybind/mgr".format(local)
             master_path = "/root/local/ceph/src/pybind"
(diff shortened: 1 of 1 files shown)


seslib/deployment.py
```diff
@@ -1210,9 +1214,37 @@ def replace_mgr_modules(self, local=None, pr=None, branch=None, repo=None, langs
                 .format(node_version, master_path, langs, npm_build))
             tools.run_sync(ssh_cmd)
 
-        for node in mgr_nodes:
+        # Fetch bin/cephadm
+        if self.settings.version not in ['nautilus', 'ses6']:
```
smithfarm, line 1218: `['ses5', 'nautilus', 'ses6']`

but maybe this would be better:

```
if self.settings.deployment_tool == 'cephadm':
```

ricardoasmarques: Done

smithfarm:
This changeset is limited to the `replace_mgr_modules` method, which I personally do not use or test.

@bk201 @tspmelo Do you care to give this a spin?

sebastian-philipp:
sounds like we duplicate cstart here?

smithfarm:
> sounds like we duplicate cpatch to some degree here?

Doesn't cpatch build a container image, though? This sesdev feature (`replace-mgr-modules`) only copies files into a running container using `podman cp`. It does not build a container image.

ricardoasmarques:
> lgtm! Just want to mentioned the cepahdm binary inside mgr container is not replaced. And that one is used by cephadm module in root mode.

Good catch, I'll fix it.

ricardoasmarques:
> > lgtm! Just want to mentioned the cepahdm binary inside mgr container is not replaced. And that one is used by cephadm module in root mode.
> 
> Good catch, I'll fix it.

Done

bk201:
> > > lgtm! Just want to mentioned the cepahdm binary inside mgr container is not replaced. And that one is used by cephadm module in root mode.
> > 
> > 
> > Good catch, I'll fix it.
> 
> Done

Tested again and it works! Thanks.

PR #393, opened 2020-07-19
deployment/status: aggregate global parameters

smithfarm:
This commit improves the "deployment status and parameters" listing
by restructuring it.

Many of the parameters that were being displayed for each node
separately are actually parameters of the deployment as a whole,
so it didn't make sense to display them for each node separately.


smithfarm:
There's a problem:

```
- ram:              0G
```

PR #398, opened 2020-07-21
Add "detail" subcommand

smithfarm:
This new subcommand provides a way to re-display the detailed
configuration report that is provided by "sesdev create".


smithfarm:
gah, already there (`sesdev show`)

PR #408, opened 2020-07-23
Change "version_os_repo_mapping" to "version_devel_repos"

smithfarm:
I never liked "version_os_repo_mapping". In our SES team parlance,
these repos are collectively referred to as "devel repos", so
the name will be more descriptive, and less confusing, if we include
the term "devel repos" in it.

Of course, it's a bit of a long-shot to say that the team is used to
referring to, e.g., "filesystems:ceph:nautilus" as the "SES6 devel repo
for OBS", but that's effectively what it is. (The OBS equivalent
of the SES6 product repo is the Leap 15.1 base repo.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -443,7 +443,7 @@ completely wrong for your use case, you can override them via
 `~/.sesdev/config.yaml`.
 
 To do this, you have to be familiar with two of sesdev's internal dictionaries:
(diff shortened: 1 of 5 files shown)


smithfarm:
As discovered by @ricardoasmarques I missed one instance of VERSION_OS_REPO_MAPPING in the code.

smithfarm:
The regression introduced by this PR is fixed by https://github.com/SUSE/sesdev/pull/419

PR #410, opened 2020-07-24
deployment: run supportconfig with 1-hour timeout

smithfarm:
Otherwise, it might run forever (or until the sesdev user hits CTRL-C,
or until the libvirtd machine powers down).

### DIFF
--- a/seslib/deployment.py
+++ b/seslib/deployment.py
@@ -963,10 +963,13 @@ class Deployment():
             self.dep_id,
             self.settings.os
             ))
-        self.ssh(name, ('supportconfig',))
+        ssh_cmd = ('timeout', '1h', 'supportconfig',)
+        self.ssh(name, ssh_cmd)
         log_handler("=> Grabbing the resulting tarball from the cluster node\n")
-        scc_exists = self.ssh(name, ('ls', '/var/log/scc*'))
-        nts_exists = self.ssh(name, ('ls', 'var/log/nts*'))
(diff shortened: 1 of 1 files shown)


smithfarm:
@tserong As part of sesdev regression testing, I run:

```
sesdev create ses5 --non-interactive --roles [master,client,openattic],[storage,mon,mgr,rgw],[stora
ge,mon,mgr,mds,nfs],[storage,mon,mgr,mds,rgw,nfs] ses5-4node
```

(which until recently was pretty reliable). During the last test run it failed in Stage 3 like so:

```
    master: Finished stage ceph.stage.3: succeeded=7/49 failed=1/49
    master: 
    master: Failures summary:
    master: ceph.mon (/srv/salt/ceph/mon):
    master:   node2.ses5-4node.test:
    master:     wait for mon: Module function cephprocesses.wait threw an exception. Exception: psutil.
NoSuchProcess process no longer exists (pid=17659)
```

Nothing shocking about that, of course, but the next thing that happened raised a question in my mind:

The next command the regression test script issues is:

```
sesdev supportconfig ses5-4node node1
```

This got to the point where it runs the ses plugin, at which point it hung:

```
  Supportconfig Plugins:                       8
    Plugin: ses...                             ^C
Aborted!
```

Would it be worth considering to put a timeout in supportconfig itself, so it doesn't hang forever in such a case? (I'll be happy to open a ticket in one of our many fine issue tracking systems if you like.)

PR #424, opened 2020-08-11
setup.cfg: fix "options.package_data" file list

smithfarm:
The "options.package_data" file list in setup.cfg determines which
template files get included in the sesdev Python package and that,
in turn, determines which files and directories get included in the
RPM package.

Commit 53c782290650ba57eff009c5dcc08d4d540a20a1 refactored the template
files but forgot to update setup.cfg accordingly.

Fixes: 53c782290650ba57eff009c5dcc08d4d540a20a1

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,11 +36,13 @@ packages =
 [options.package_data]
 seslib =
     templates/*.j2
-    templates/deepsea/*.j2
     templates/caasp/*.j2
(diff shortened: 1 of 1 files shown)


smithfarm:
The currently shipping RPM (1.6.0) is broken and this (hopefully) is the fix so I'll go ahead and merge it now.

PR #427, opened 2020-08-12
single node CaaSP cluster

denisok:
add functionality to deploy single node CaaSP cluster with ability to
deploy Rook on it

Signed-off-by: Denis Kondratenko <denis.kondratenko@suse.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -390,7 +390,7 @@ To create CaaSP k8s cluster that has loadbalancer, 2 worker nodes and master:
 $ sesdev create caasp4
 ```
 
-By default it just creates and configures CaaSP cluster and workers don't have any disks.
+By default it just creates and configures CaaSP cluster and workers don't have any disks if there is no `--deploy-ses` option.
 
 To create workers with disks and without loadbalancer role:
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -390,16 +390,27 @@ To create CaaSP k8s cluster that has loadbalancer, 2 worker nodes and master:
...
 ```
 
-To deploy Rook on that cluster use `--deploy-ses` option, also be aware about minimum requirement for the disk sizes (>5G).
+To deploy Rook on that cluster use `--deploy-ses` option, default disk size would be 8G, num of disks per node 2:
 
+```
+$ sesdev create caasp4 --deploy-ses
+```
+
+To create single node cluster use `--single-node` option, for example this creates 1 node with 2 disks (8G) and deploy Rook:
```
smithfarm, line 407: `--single-node` should always give 4 disks. If that's not the case with caasp4 --deploy-ses, then we should fix it

denisok: https://github.com/SUSE/sesdev/blob/master/sesdev/__init__.py#L876
it is 2.

Issue #430, opened 2020-08-14
sesdev tries to install python2-pip on Python3 system


votdev:
When running `sesdev replace-mgr-modules ses7-mini --pr <ID>` the package `python2-pip` is tried to be installed, but that fails because SES7 is Python3 only.

```
 ~/git/sesdev   master ✔  $ sesdev replace-mgr-modules ses7-mini --pr 36611
Fetching PR '36611' from 'ceph'...
Disabling modules...

Fetching...
Building...
Command '['ssh', 'root@192.168.121.98', '-i', '/home/vot/.sesdev/ses7-mini/keys/sesdev', '-o', 'IdentitiesOnly yes', '-o', 'StrictHostKeyChecking no', '-o', 'UserKnownHostsFile /dev/null', '-o', 'PasswordAuthentication no', "cd ~/ &&                 zypper -n in python2-pip python3-pip &&                 pip install nodeenv &&                 nodeenv env --node=10.18.1 --force &&                 . ~/env/bin/activate &&                 cd /root/remote/ceph/src/pybind/mgr/dashboard/frontend &&                 export DASHBOARD_FRONTEND_LANGS='en-US' &&                 export NG_CLI_ANALYTICS=false &&                 npm ci --unsafe-perm &&                 npm run build &&                 rm -rf node_modules"]' failed: ret=104 stderr:
b"Warning: Permanently added '192.168.121.98' (ECDSA) to the list of known hosts.\r\nKeine Anbieter von 'python2-pip' gefunden.\n"
```

smithfarm:
This was introduced by https://github.com/SUSE/sesdev/pull/24 and I've commented there, but the simple fix would be to drop `python2-pip` from this line:

```
zypper -n in python2-pip python3-pip && \
```

I'll leave it to @tspmelo to fix, though, since it's his code.

tspmelo:
We still need that for nautilus based clusters.
The solution should be to check what kind of cluster we have and install one or the other.

It seems this only fails when all other packages are already installed, otherwise zypper doesn't complain, and must have escaped my testing.

I will be on vacation next week. Will work on it after it.

smithfarm:
> We still need that for nautilus based clusters.

@tspmelo Why do you think Python 2 is needed in Nautilus/SES6, though? While it is true that `zypper in python2-pip` will bring in Python 2 if it's available in the repos, it should not be needed to run Ceph, except on SES5.

[...]

PR #431, opened 2020-08-14
replace-mgr-modules: refrain from introducing Python 2

smithfarm:
We are already ensuring that Python 2 will not be present in the system,
so installing "python2-pip" here will actually pull in the entire Python
2 stack. Also, Python 2 is only available in SLE-15-SP2 via the "Python
Legacy" module which sesdev does not add to the deployment.

Fixes: https://github.com/SUSE/sesdev/issues/430


smithfarm:
@tspmelo I won't merge this until you get back.

@votdev Can you try this? Does it resolve the issue you reported in #430 ?

tspmelo (approved):
lgtm

PR #433, opened 2020-08-17
README: problems accessing dashboard on remote sesdev

smithfarm:
Add new section so I don't have to re-invent the wheel every time I need
to access the dashboard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ The Jenkins CI tests that `sesdev` can be used to deploy a single-node Ceph
    * [sesdev destroy reported an error](#sesdev-destroy-reported-an-error)
    * ["Failed to connect socket" error when attempting to use remote libvirt server](#failed-to-connect-socket-error-when-attempting-to-use-remote-libvirt-server)
    * [mount.nfs: Unknown error 521](#mountnfs-unknown-error-521)
+   * [Problems accessing dashboard on remote sesdev](#problems-accessing-dashboard-on-remote-sesdev)
 * [Contributing](#contributing)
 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1053,6 +1054,60 @@ If this doesn't report back with `active`, please consider running:
...
+```
+
+it will choose an IP address essentially at random. It's highly probable that
+your remote sesdev machine has multiple IP addresses and sesdev has chosen the
+"wrong" one (i.e., one that is not accessible from the laptop where the browser
+is running).
+
+However, even when specifying `--local-address CORRECT_IP_ADDRESS`, it still
+might not work if there are other dashboard instances (sesdev or bare metal)
+listening on the default port 8443 on the machine.
```
asettle, line 1081: (or 8080 when SSL is disabled, but I highly doubt you need to mention that)

README.md
```diff
@@ -1053,6 +1054,60 @@ If this doesn't report back with `active`, please consider running:
...
+
+#### Resolution
+
+First, you have to be really sure which IP address of the sesdev machine is
+accessible from what I'm calling the "laptop" (the machine where the browser is
```
asettle, line 1086: accessible from your local machine - no? Sounds simpler?

README.md
```diff
@@ -1053,6 +1054,60 @@ If this doesn't report back with `active`, please consider running:
...
+"wrong" one (i.e., one that is not accessible from the laptop where the browser
+is running).
+
+However, even when specifying `--local-address CORRECT_IP_ADDRESS`, it still
+might not work if there are other dashboard instances (sesdev or bare metal)
+listening on the default port 8443 on the machine.
+
+#### Resolution
+
+First, you have to be really sure which IP address of the sesdev machine is
```
asettle, line 1085: This is confusing. In my head, I auto-translated "which" to "that".

README.md
```diff
@@ -1053,6 +1054,60 @@ If this doesn't report back with `active`, please consider running:
...
+```
+
+This will tell you the node where the dashboard is running. Now, do:
+
+```
+sesdev tunnel DEP_ID \
+    --node NODE \
+    --remote-port 8443 \
+    --local-address CORRECT_IP_ADDRESS \
+    --local-port SOME_HIGH_NUMBERED_PORT
```
asettle, line 1102: This might be a bit confusing. A note to explain that you're suggesting using _any_ high numbered port might be helpful for future

README.md
```diff
@@ -1053,6 +1054,60 @@ If this doesn't report back with `active`, please consider running:
...
+```
+
+Then point the browser at `https://CORRECT_IP_ADDRESS:SOME_HIGH_NUMBERED_PORT`.
```
asettle, line 1105: Also don't forget that the output will not return "https://BLAH:BLAH" it will only return "BLAH:BLAH" - the https:// needs to be appended - also means that the output in the console is not clickable (as it is in the normal output)

smithfarm:
@asettle Thanks for the careful review!

PR #435, opened 2020-08-19
vet_configuration: vet caasp4 roles more carefully

smithfarm:
This PR is a mixed bag of `caasp4` cleanups and fixes:

1. caasp4 clusters must not have more than one role per machine, which might not be clear to all users
2. the regex for counting worker nodes would only work up to 10 worker nodes, but there's no reason why it couldn't handle more
3. finally, I found that we were adding disks even when the user doesn't need them - add a commit to address that

### DIFF
--- a/README.md
+++ b/README.md
@@ -391,7 +391,9 @@ To create CaaSP k8s cluster that has loadbalancer, 2 worker nodes and master:
 $ sesdev create caasp4
 ```
 
-By default it just creates and configures CaaSP cluster and workers don't have any disks if there is no `--deploy-ses` option.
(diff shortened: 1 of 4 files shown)


denisok (approved):
all good, but please update README with this, so ppl wouldn't wonder what is the logic behind.

smithfarm:
@denisok I added a commit expanding the README, and I also expanded the exception like so:

```
            "Multiple roles per machine detected. This is not allowed in CaaSP "
            "clusters. For a single-node cluster, use the --single-node option "
            "or --roles=\"[master]\" (in this special case, the master node "
            "will function also as a worker node)"
```

denisok (approved):

PR #437, opened 2020-08-19
deepsea: add ganesha roles to policy.cfg

smithfarm:
One fine day, we noticed that sesdev was not adding ganesha roles to
policy.cfg in {ses5,nautilus,ses6} clusters.


supriti (approved):

smithfarm:
Actually, this isn't as bad as it looks. We just got confused because the `ganesha` role got renamed to `nfs` by 26d321e3fdae8c4d14b028124b47125ce23fffa0, but I had forgotten that this was done for deployment versions that use DeepSea as a deployment tool. This is confusing because DeepSea also has a concept of "roles" and it only recognizes "ganesha", not "nfs".

Taking corrective action in #442 ...

PR #443, opened 2020-08-20
ceph-salt: enable user to control which nodes get "admin" role

smithfarm:
The documentation states that the only node that **must** have an
"admin" role is the bootstrap node. On all other nodes, including the
master node, "admin" is optional.

I realized that our previous approach, which was to apply the "admin"
role to all nodes without exception, might not be ideal for realistic
testing. According to the SES7 manual, this configuration is not
recommended.

So, this commit makes it possible for the user to specify which nodes
get the "admin" role when deploying {octopus,ses7,pacific} clusters, 
with the caveat that then "admin" role must be given on nodes that 
have the "bootstrap" and/or the "master" role. (On the "master" node,
"admin" is not strictly required, but it seems safe to assume that most
real-life deployments will have it, since it's natural to want to run ceph
CLI commands on the master node.)

### DIFF
--- a/contrib/standalone.sh
+++ b/contrib/standalone.sh
@@ -282,7 +282,7 @@ fi
 if [ "$SES7" ] ; then
     sesdev --verbose box remove --non-interactive sles-15-sp2
     run_cmd sesdev --verbose create ses7 --dry-run
-    run_cmd sesdev --verbose create ses7 --non-interactive --roles "[master,storage,mon,mgr]" --qa-test ses7-mini
+    run_cmd sesdev --verbose create ses7 --non-interactive --roles "[admin,master,bootstrap,storage,mon,mgr]" --qa-test ses7-mini
     run_cmd sesdev --verbose destroy --non-interactive ses7-mini
     run_cmd sesdev --verbose create ses7 --non-interactive "${CEPH_SALT_OPTIONS[@]}" --single-node --qa-test ses7-1node
     run_cmd sesdev --verbose destroy --non-interactive ses7-1node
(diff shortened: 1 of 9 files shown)


smithfarm:
Probably we need to give "admin" role to the master node by default.

smithfarm:
This makes the sesdev deployment a little more realistic. By default, `admin` role is mandatory only for nodes that have `bootstrap` and/or `master` role. Also, the user now has control over which nodes get the `admin` role.

ricardoasmarques (changes requested):
I've created a `pacific` env by running:

```
sesdev create pacific --ceph-salt-repo https://github.com/ceph/ceph-salt.git --ceph-salt-branch master pacific
```

Then I tried to SSH into that cluster, but it fails:

```
# sesdev ssh pacific
Node 'master' does not exist in this deployment

# sesdev list
Found 1 deployment:

+---------+---------+---------+---------------------+
| ID      | Version | Status  | Nodes               |
+---------+---------+---------+---------------------+
| pacific | pacific | running | admin, node1, node2 |
+---------+---------+---------+---------------------+
```

smithfarm:
@ricardoasmarques Thats strange and unexpected, and I cannot reproduce it here. Somehow this piece of code (which I added a long time ago for backward compatibilty) is being triggered in your case:

https://github.com/SUSE/sesdev/blob/master/seslib/deployment.py#L280-L283

Can you capture the full output of your `sesdev create` command in a file and post it somewhere?

smithfarm:
@ricardoasmarques I added a commit that should fix the issue for you. It's weird, though - I can't reproduce it.

ricardoasmarques (approved):
LGTM

PR #445, opened 2020-08-24
replace-mgr-modules: Fix nautilus mgr copy

tspmelo:
Its not possible to use rsync between 2 remotes.
This code tries to do the same as the block after it,
so we can simply remove it.

Signed-off-by: Tiago Melo <tmelo@suse.com>

### DIFF
--- a/seslib/deployment.py
+++ b/seslib/deployment.py
@@ -1375,13 +1375,6 @@ deployment might not be completely destroyed.
             # Copy mgr modules
             if self.settings.version in ['nautilus', 'ses6']:
                 print("Copying mgr modules to node {}".format(node))
-                self.rsync(
-                    '{}:{}/mgr'.format(master_node, master_path),
-                    "{}:/usr/share/ceph/".format(node),
-                    excludes=['.git', '.tox', '.idea', 'venv', 'build',
-                              'node_modules', 'cypress'],
-                    recurse=True)
-
                 ssh_cmd = self._ssh_cmd(master_node)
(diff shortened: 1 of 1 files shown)


smithfarm (approved):
Why is it not possible to rsync between two nodes? I can ssh, e.g., from `node1` to `node2`, so rsync should also be possible.

tspmelo:
> Why is it not possible to rsync between two nodes? I can ssh, e.g., from `node1` to `node2`, so rsync should also be possible.

Didn't investigate it, just saw the message that was displayed in the console and assume it was a limitation.

smithfarm:
> just saw the message that was displayed in the console

Would be nice to know what message that is. I'll try to reproduce it.

tspmelo:
This is what is displayed when trying to copy in nautilus:
```
Copying mgr modules to node node1
The source and destination cannot both be remote.
rsync error: syntax or usage error (code 1) at main.c(1436) [Receiver=3.2.3]
Copying mgr modules to node node2
The source and destination cannot both be remote.
rsync error: syntax or usage error (code 1) at main.c(1436) [Receiver=3.2.3]
Copying mgr modules to node node3
The source and destination cannot both be remote.
rsync error: syntax or usage error (code 1) at main.c(1436) [Receiver=3.2.3]
Enabling modules...
```

smithfarm (approved):

smithfarm:
@tspmelo Sorry this took me so long to review. As you say, the `scp` that follows immediately after the `rsync` is copying more-or-less the same set of files, and the `rsync` returns the error you quoted - so your fix makes sense. I guess my hesitation was because I don't understand how this worked in the first place.

Issue #37, opened 2019-12-20
Ability to rotational/non-rotational disks ("--ssd" option to "sesdev create")


smithfarm:
I heard a rumor - based on https://bugzilla.redhat.com/show_bug.cgi?id=1498042#c14 - that QEMU/libvirt might offer the option of specifying rotational/non-rotational flag for disks. If this is possible, we should update sesdev to support it.

Of course, this depends on whether *vagrant* supports passing that to QEMU/libvirt.

smithfarm:
The vagrant-libvirt documentation seems to indicate there is no way to specify "non-rotational":

https://github.com/vagrant-libvirt/vagrant-libvirt#additional-disks

smithfarm:
Ah, but wait:

"In absence of information to the contrary, Linux guests assume that all block devices are backed by rotational storage. The guest OS admin can override this by setting the 'rotational' flag in sysfs."

smithfarm:
The standard use case for a non-rotational disk would be for use as bluestore db/wal for the rotational disks on the node. So, we could add a `--ssd` parameter to `sesdev create` which would set the rotational flag for one of the additional disks (for all nodes that have such disks).

PR #448, opened 2020-08-28
If --ssd option is given, make first additional disk non-rotational

smithfarm:
This commit adds support for simulating mixed SSD/HDD clusters. For the
time being, we keep it simple:

1. if --ssd is not given, all disks are rotational (as before)
2. if --ssd is given, the first additional disk on each node is set to
   "non-rotational", thereby simulating an SSD

In other words, this commit does not add support for complicated
configurations like

1. multiple SSDs per node
2. clusters where disk HW configurations vary from node to node

Fixes: https://github.com/SUSE/sesdev/issues/37

### DIFF
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -144,7 +144,9 @@ def common_create_options(func):
         click.option('--synced-folder', type=str, default=None, multiple=True,
                      help='Set synced-folder to be mounted on the master node. <str:dest>'),
         click.option('--dry-run/--no-dry-run', is_flag=True, default=False,
-                     help='Dry run (do not create any VMs)')
+                     help='Dry run (do not create any VMs)'),
+        click.option('--ssd', is_flag=True, default=False,
+                     help='On VMS with additional disks, make one disk non-rotational')
     ]
     return _decorator_composer(click_options, func)
 
@@ -445,6 +447,7 @@ def _gen_settings_dict(
         scc_pass=None,
         scc_user=None,
         single_node=None,
+        ssd=None,
         stop_before_ceph_orch_apply=None,
         stop_before_ceph_salt_apply=None,
         stop_before_ceph_salt_config=None,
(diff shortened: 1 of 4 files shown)


bk201:
Thanks for adding this feature, it's useful for testing WAL/DB devices.

I used the same hack to emulate SSDs before, but somehow the rotational flag is always reset back to 1 after performing `ceph orch device ls --refresh` (Centos 8, upstream).

So later I use the help from qemu, but I beilive this requires newer version of qemu.
The following the snippet from my Vagrant file:

```
    config.vm.define "mgr#{i}" do |mgr|
      mgr.vm.hostname = "mgr#{i}"
      mgr.vm.provider :libvirt do |libvirt|
        libvirt.storage :file, :size => '10G', bus: "sata", serial: "mgr#{i}-1-ssd"
        libvirt.storage :file, :size => '10G', bus: "sata", serial: "mgr#{i}-2-ssd"
        libvirt.storage :file, :size => '20G', bus: "sata", serial: "mgr#{i}-3-hdd"
        libvirt.storage :file, :size => '20G', bus: "sata", serial: "mgr#{i}-4-hdd"
        libvirt.qemuargs :value => "-set"
        libvirt.qemuargs :value => "device.sata0-0-1.rotation_rate=1"
        libvirt.qemuargs :value => "-set"
        libvirt.qemuargs :value => "device.sata0-0-2.rotation_rate=1"
      end
    end
```

References:
* https://blog.christophersmart.com/2019/12/18/kvm-guests-with-emulated-ssd-and-nvme-drives/
* https://github.com/vagrant-libvirt/vagrant-libvirt#custom-command-line-arguments-and-environment-variables
* https://unix.stackexchange.com/questions/235414/libvirt-how-to-pass-qemu-command-line-args

smithfarm:
Thanks, @bk201, that's exactly the information I was hoping to find! I can see there are currently two versions of qemu at play:

1. Leap 15.1 has qemu version 3.1.1.1
2. Leap 15.2 has qemu version 4.2.1

Are either of these new enough for this feature?

bk201:
> Do you know which qemu version added this feature?

I don't have an answer without further digging. The best bet is to test it in Leap 15.1, and if it works, the others should work too. But how about users not using OpenSUSE? It would be nice if qemu have a feature indicattion flag.

PR #452, opened 2020-09-03
ceph_salt_deployment: prevent Vagrant from deleting master node

smithfarm:
As of some recent version of Vagrant, it started removing the master
domain (i.e., destroying the master node) whenever the provisioning
script exits with a code other than "0".

Deleting the master node makes it impossible to investigate why the
script failed.

I can't think of any good reason for Vagrant/vagrant-libvirt to behave
like this. It's a recent development. I don't know how to turn the
behavior off, but I can at least work around it by preventing the
deployment script from returning any exit code other than 0.


smithfarm:
This doesn't work. We need to trap the signal produced by `set -e` when something fails and issue `exit 0`.

Sigh.

smithfarm:
@LenzGr This PR is still in flux. I'll have an update later today.

PR #469, opened 2020-09-11
contrib/standalone.sh: remove config.yaml even when failing

smithfarm:
When a test failed, the config.yaml that was created for testing
purposes was being left behind.


smithfarm:
Follow-on fix: 402a0a39adc80e489e3d2218d315fe111fc9ad76

PR #472, opened 2020-09-12
provision.sh: trap CTRL-C as well as errors

smithfarm:
Vagrant 2.2.10 has a regression where it removes the master node
(immediately) when the provisioner returns a non-zero exit status.

Return zero even on CTRL-C and, as a bonus, show the line number the
script was at when CTRL-C was pressed.

### DIFF
--- a/seslib/templates/provision.sh.j2
+++ b/seslib/templates/provision.sh.j2
@@ -14,6 +14,7 @@ function err_report {
 }
 
 # ensure Vagrant doesn't destroy the master node when the script fails
+trap 'err_report $LINENO' INT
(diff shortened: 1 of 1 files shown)


smithfarm:
The first CTRL-C only pauses the output. When CTRL-C is pressed a second time, Vagrant issues the following message:

```
Aborted!
```

and returns 1. With Vagrant 2.2.10 this will presumably cause the master VM to be immediately deleted.

In any case, this fix doesn't work, possibly due to https://github.com/hashicorp/vagrant/issues/9850 (?)

PR #453, opened 2020-09-04
Use the official boxes from VagrantCloud when available

dcermak:

PR #476, opened 2020-09-13
Support official Vagrant boxes from Vagrant Cloud

smithfarm:
This is a continuation of the work started by @dcermak in #453 

The idea is to use official Vagrant Boxes (from Vagrant Cloud) when available.

There is some complexity involved, because the official boxes have hard-coded box names, while `sesdev` expects its boxes to **not** have a box name. To overcome this, this PR aspires to implement a box aliasing system.

### DIFF
--- a/seslib/deployment.py
+++ b/seslib/deployment.py
@@ -36,10 +36,10 @@ from .exceptions import \
                         ServicePortForwardingNotSupported, \
                         SubcommandNotSupportedInVersion, \
                         SupportconfigOnlyOnSLE, \
-                        VagrantBoxDoesNotExist, \
                         VagrantSshConfigNoHostName, \
                         VersionOSNotSupported, \
-                        UniqueRoleViolation
+                        UniqueRoleViolation, \
+                        UnsupportedVMEngine
 from .log import Log
 from .node import Node, NodeManager
 from .settings import Settings, SettingsEncoder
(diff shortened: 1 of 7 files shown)


seslib/deployment.py
```diff
@@ -525,37 +525,40 @@ def _generate_vagrantfile(self):
         return scripts
 
     def save(self):
+        #
+        # by "scripts", we mean the Vagrantfile itself plus one provisioning
+        # script for each node
         scripts = self._generate_vagrantfile()
         key = RSA.generate(2048)
```
dcermak, line 535: I know that this is unrelated, but why not use ed25519 instead? It is faster and at this point universally supported. Also, 2048 bit RSA is not really up to date from a security standpoint.

dcermak (commented):
Note that I am not really familiar with the sesdev code base, so take this with a grain of salt. Overall the changes look good to me and if the tests are passing, I have no objection to merging this.

seslib/deployment.py
smithfarm: sesdev is not particularly concerned about the security of the SSH keys it generates. Its purpose is to deploy Ceph clusters on libvirt VMs for development/testing purposes. These clusters are assumed to be short-lived and not SSHable from the outside.

Would using ed25519 instead of RSA introduce a new dependency? Currently we are getting RSA from:

```
Requires:       python3-pycryptodomex >= 3.4.6
```

dcermak: Unfortunately pycryptodomex does not support ed25519 keys, as they only provide the NIST curves and Daniel Bernstein's ed25519 is not one of them.

PR #481, opened 2020-09-15
Add new msgr2 secure options to sesdev

matthewoliver:
In order to deploy a cluster with messger v2 encryption we need to
prioritise 'secure' over 'crc' mode or only provide secure mode.
The specific ceph options are:
 - ms_cluster_mode
 - ms_service_mode
 - ms_client_mode

By default all of these are set to 'crc secure' which prioritises 'crc'
over full encryption.

NOTE: there are 3 mon specific options to the above (ms_mon_*_mode) that
by defaults already use 'secure crc' so they'll prefer secure mode first
over crc between the mon nodes when they speak msgr2. Which is awesome.
I could add an option to force mon "secure" too.

This patch adds 2 new flags that control msgr2 secure mode:

 --msgr2-secure-mode: This sets the above 3 options to just 'secure'.
 --msgr2-prefer-secure: This changes the order to 'secure crc' so secure
is prefered over crc.

These only effect msgr2, so anything talking msgr1 (like the kernel
client) will be unencrypted.

Signed-off-by: Matthew Oliver <moliver@suse.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ The Jenkins CI tests that `sesdev` can be used to deploy a single-node Ceph
          * [octopus from filesystems:ceph:octopus&#8203;:upstream](#octopus-from-filesystemscephoctopusupstream)
          * [ses7 from Devel:Storage:7.0](#ses7-from-develstorage70)
          * [ses7 from Devel:Storage:7.0:CR](#ses7-from-develstorage70cr)
+      * [With wire encryption](#with-wire-encryption)
       * [Deploying non-SUSE environments](#deploying-non-suse-environments)
          * [Ubuntu "Bionic Beaver" 18.04](#ubuntu-bionic-beaver-1804)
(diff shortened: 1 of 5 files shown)


smithfarm (approved):
Thanks, @matthewoliver ! It looks innocuous enough, but I'll run it through `contrib/standalone.sh` before merging it.

smithfarm (changes requested):
@matthewoliver Could you add a section `On-the-wire encryption (msgr2)` to README.md ?

You have a nice write-up in the commit message that would be a good start. But it doesn't need to be too technical!

smithfarm:
@matthewoliver I'm also curious how one would verify that the wire communications are really encrypted. Any thoughts on that?

smithfarm:
I went ahead and added the README.md verbiage so this can be merged without further delay.

PR #484, opened 2020-09-16
zypper: install rbd-nbd on all Leap 15.2/SLE-15-SP2 nodes

smithfarm:
On SLE-15-SP2, rbd-nbd should now be a part of the Base System module.

References: https://jira.suse.com/browse/SLE-11021


smithfarm:
Right now this causes the ses7 deployment to break, but that should be fixed soon by adding the rbd-nbd package to the SES7 Media.

PR #489, opened 2020-09-17
Update pacific repos

ricardoasmarques:
Use pacific repos instead of master.

### DIFF
--- a/seslib/constant.py
+++ b/seslib/constant.py
@@ -21,7 +21,7 @@ class Constant():
     IMAGE_PATHS = {
         'ses7': 'registry.suse.de/devel/storage/7.0/containers/ses/7/ceph/ceph',
         'octopus': 'registry.opensuse.org/filesystems/ceph/octopus/images/ceph/ceph',
-        'pacific': 'registry.opensuse.org/filesystems/ceph/master/upstream/images/ceph/ceph',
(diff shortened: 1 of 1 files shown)


seslib/constant.py
```diff
@@ -252,11 +252,11 @@ class Constant():
         },
         'pacific': {
             'leap-15.2': [
-                'https://download.opensuse.org/repositories/filesystems:/ceph:/master:/upstream/'
+                'https://download.opensuse.org/repositories/filesystems:/ceph:/pacific/'
                 'openSUSE_Leap_15.2'
             ],
             'tumbleweed': [
-                'https://download.opensuse.org/repositories/filesystems:/ceph:/master:/upstream/'
+                'https://download.opensuse.org/repositories/filesystems:/ceph:/pacific:/upstream/'
```
smithfarm, line 259: I would suggest to use `filesystems:ceph:pacific` for Tumbleweed as well.

smithfarm: In a follow-on PR we can add a `master` version target that will consume the `filesystems:ceph:master:upstream` repo/container

ricardoasmarques: Fixed

smithfarm (approved):

Issue #396, opened 2020-07-20
Provide a way to connect two sesdev clusters


trociny:
If I deploy two sesdev clusters the network connection between them is forbidden on iptables level. Sometimes we need to connect the clusters though (I needed it for testing RBD mirroring. And I can imagine it could be useful for testing RGW multisite or CephFS snapshots sync).

I found a way to just manually tweak the iptables rules after the clusters are deployed. Though I believe there is a nicer way or it could be implemented (documented).

smithfarm:
https://github.com/ldx/python-iptables

PR #497, opened 2020-09-19
[DNM] Implement "sesdev link" by adding rules to LIBVIRT_FWI chain

smithfarm:
From the feature request:

"If I deploy two sesdev clusters the network connection between them is
forbidden on iptables level. Sometimes we need to connect the clusters
though (I needed it for testing RBD mirroring. And I can imagine it
could be useful for testing RGW multisite or CephFS snapshots sync)."

Fixes: https://github.com/SUSE/sesdev/issues/396


smithfarm:
The LIBVIRT_FWI chain is maintained by libvirt, so adding rules to it is probably not a good idea.

PR #502, opened 2020-09-20
Implement "sesdev status" subcommand

smithfarm:

---

TO DO:

- [x] README.md
- [x] contrib/standalone.sh
- [x] `--format json`


asettle (approved):
Oh I like this, I always type "sesdev status" before I remember it's "list"

PR #522, opened 2020-10-06
cephadm/deployment_day_2: specify FQDN when appropriate

smithfarm:
When the user provides --fqdn, "hostname" will return the FQDN and, in
that case, we'd better give cephadm FQDNs if we want it to continue
being nice to us.


denisok (approved):

smithfarm:
Another approach would be to use a shell variable, like this:

    MY_HOSTNAME="$(hostname)"

but that introduces its own specific set of requirements. So I'd rather be explicit like this.

Issue #136, opened 2020-03-05
octopus/ses7: Allow Ceph clusters to be deployed with AppArmor completely disabled


smithfarm:
Octopus deployments currently have AppArmor enabled by default.

Past experience with testing ceph deployments has shown that AppArmor, when enabled, can potentially be a big headache, because it typically causes errors that are difficult to link to AppArmor.

Clearly, an easy way to turn off AppArmor is needed -- i.e., an`--apparmor/--no-apparmor` option to `sesdev create`.

However, this option must ensure that AppArmor is not running anywhere. It is not sufficient to disable it on the nodes themselves, because it might be running in the containers.

smithfarm:
Blocked by https://github.com/ceph/ceph-salt/issue/12

smithfarm:
https://bugzilla.suse.com/show_bug.cgi?id=1165821 is now fixed with the advent of podman 1.8.0

Hopefully, that means https://github.com/ceph/ceph/pull/33850 will now be reverted and work on https://github.com/ceph/ceph-salt/issues/12 can go forward.

smithfarm:
Right now this is blocked because it's not clear how to disable AppArmor completely in SLE-15-SP2.

PR #543, opened 2020-10-14
Implement --no-apparmor option

smithfarm:
Until now, if AppArmor was active in the VM image, sesdev gave you no
way of deactivating it.

Fixes: https://github.com/SUSE/sesdev/issues/136


smithfarm:
CI failure is unrelated.

PR #547, opened 2020-10-21
deepsea/nautilus_pre_stage_0.sh: refrain from patching DeepSea

smithfarm:
For a long time, sesdev was patching the DeepSea code. This was needed
because, in sesdev clusters, one of the hosts was called "admin" and the
MDS doesn't like this.

Now that sesdev is no longer calling hosts "admin", the patch can be
dropped.


smithfarm:
@jan--f I'm not sure why Ricardo thought this patch was needed... testing without the patch to (hopefully) find out.

jan--f (approved):

PR #552, opened 2020-10-29
constant,zypper: use download.nue.suse.com instead of download.suse.de

smithfarm:
download.nue.suse.com is accessible to a wider range of people than
download.suse.de, but otherwise equivalent.


jan--f (approved):
:+1:

smithfarm:
Thanks, Jan - I had forgotten about `download.nue.suse.com`. Too bad there's no `registry.nue.suse.com` yet...

PR #555, opened 2020-11-12
deployment_day_2.sh: create CephFS using documented procedure

smithfarm:
It came to my attention that sesdev was using an unsupported
and undocumented procedure ("ceph fs volume create") to create
the CephFS.


jan--f (approved):
:+1:

Issue #563, opened 2020-11-25
image_paths override in config.yaml broken since introduction of image_paths_{product,devel}


tserong:
https://github.com/SUSE/sesdev/pull/561 replaced `image_paths` in settings.py with `image_paths_product` and `image_paths_devel`. Likewise `Constant.IMAGE_PATHS` became `Constant.IMAGE_PATHS_PRODUCT` and `Constant.IMAGE_PATHS_DEVEL`. However, `Settings._load_config_file` still uses `image_paths` and `Constant.IMAGE_PATHS`:
https://github.com/SUSE/sesdev/blob/f9201cab5b72b66e383a5dbcf7b7636328cc8eae/seslib/settings.py#L434
i.e. the override in config.yaml is still expected to only be `image_paths`. This causes `sesdev create` to fail with:

```
...
    __fill_in_config_tree('image_paths', Constant.IMAGE_PATHS)
AttributeError: type object 'Constant' has no attribute 'IMAGE_PATHS'
```

The easy fix is:

```diff --git a/seslib/settings.py b/seslib/settings.py
index 216b6da..d7a9b48 100644
--- a/seslib/settings.py
+++ b/seslib/settings.py
@@ -431,7 +431,8 @@ class Settings():
         assert isinstance(config_tree, dict), "yaml.load() of config file misbehaved!"
         __fill_in_config_tree('os_repos', Constant.OS_REPOS)
         __fill_in_config_tree('version_devel_repos', Constant.VERSION_DEVEL_REPOS)
-        __fill_in_config_tree('image_paths', Constant.IMAGE_PATHS)
+        __fill_in_config_tree('image_paths_devel', Constant.IMAGE_PATHS_DEVEL)
+        __fill_in_config_tree('image_paths_product', Constant.IMAGE_PATHS_PRODUCT)
         __fill_in_config_tree('version_default_roles', Constant.ROLES_DEFAULT_BY_VERSION)
         return config_tree

```

I was going to submit a PR with that, but then realised README.md has a bunch of example config.yaml files using the old `image_paths` form, and I wanted to check first to make sure this was the correct approach, before rewriting those examples too.

smithfarm:
So, yeah, it looks like #561 did "omit" (run roughshod over?) the necessary changes to the `__fill_in_config_tree` calls and `README.md`. I don't think you're missing anything.

PR #564, opened 2020-11-25
Handle image_paths_{product,devel} overrides in config.yaml

tserong:
Fixes: https://github.com/SUSE/sesdev/issues/563

### DIFF
--- a/README.md
+++ b/README.md
@@ -644,16 +644,16 @@ This would cause the zypper repo to be added at priority 96.
 
 In Ceph versions "octopus" and newer, the Ceph daemons run inside containers.
 When the cluster is bootstrapped, a container image is downloaded from a remote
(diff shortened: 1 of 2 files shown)


smithfarm (approved):
We might need more detail in `README.md` at some point - not sure about that - but this information should be enough . . . as long as the user does not specify the `--product` option.

The thing is: repos are not simple and it's impossible to "abstract them away". Containers were supposed to do that, but they failed to do so, and now things are even *more* complicated.

tserong:
> We might need more detail in `README.md` at some point - not sure about that - but this information should be enough . . . as long as the user does not specify the `--product` option.

Yeah... But let's see how we go. I kinda didn't want to rework a whole bunch of explanation right now with this PR :-)

PR #565, opened 2020-11-27
seslib/constant.py: add SES7 Update repo

smithfarm:
Now that SES7 is a released product, it is receiving maintenance updates and we
will need to always have the SES7 Update repo available.


colder-is-better:
That's very convenient -- thanks!

PR #566, opened 2020-12-02
provision: enable coredumps

smithfarm:
It came to my attention that coredump capturing might not be enabled by default
in sesdev clusters.

### DIFF
--- a/seslib/templates/provision.sh.j2
+++ b/seslib/templates/provision.sh.j2
@@ -16,6 +16,9 @@ function err_report {
 # ensure Vagrant doesn't destroy the master node when the script fails
 trap 'err_report $LINENO' ERR
 
+# do not limit coredump size
+ulimit -c unlimited
+
 ls -lR /home/vagrant
 
 # populate /etc/hosts
@@ -96,6 +99,23 @@ else
 fi
 aa-status || true
 {% endif %}{# apparmor #}
+
+{% if os != 'sles-12-sp3' %}
+source /etc/os-release
+if [[ "$ID_LIKE" =~ "suse" ]] ; then
+    # enable coredump capturing
+    sysctl -w kernel.core_pattern="|/usr/lib/systemd/systemd-coredump %P %u %g s %t %c %e"
+    sysctl kernel.core_pattern
+    cat <<EOF >>/etc/systemd/coredump.conf
+ProcessSizeMax={{ ram }}G
+ExternalSizeMax={{ ram }}G
+JournalSizeMax={{ ram }}G
+EOF
+    cat /etc/systemd/coredump.conf
+    systemctl enable systemd-coredump.socket
+    systemctl start systemd-coredump.socket
+fi
+{% endif %}{# os != 'sles-12-sp3' #}
 
 # deployment state machine
 
--- a/qa/common/common.sh
+++ b/qa/common/common.sh
@@ -1303,3 +1303,59 @@ function alertmanager_smoke_test {
 function node_exporter_smoke_test {
     _monitoring_smoke_test "node-exporter" "title.Node.Exporter"
 }
+
+function core_dump_test {
+    local skipped
+    local succeeded
+    local osd_0_ps
+    local osd_0_pid
+    echo "WWWW: core_dump_test"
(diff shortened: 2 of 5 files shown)


seslib/templates/provision.sh.j2
```diff
@@ -97,6 +100,10 @@ fi
 aa-status || true
 {% endif %}{# apparmor #}
 
+# enable coredump capturing
+systemctl enable systemd-coredump.socket
+systemctl start systemd-coredump.socket
+
```
denisok, line 106: should you also do:
`sysctl -w kernel.core_pattern="|/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %e"`
?

seslib/templates/provision.sh.j2
```diff
@@ -97,6 +100,15 @@ fi
...
 {% endif %}{# apparmor #}
 
+# enable coredump capturing
+cat <<EOF >>/etc/systemd/coredump.conf
+ExternalSizeMax=10G
+JournalSizeMax=10G
+EOF
+cat /etc/systemd/coredump.conf
+systemctl enable systemd-coredump.socket
+systemctl start systemd-coredump.socket
```
ddiss, line 110: Aren't `socket` services supposed to start automatically once an IPC/network client attempts to connect to the socket? If so, is the `start` here necessary?
Looks fine to me otherwise (although those SizeMax values are rather large)

djz88: Default values in `/etc/systemd/coredump.conf` are:
```
#Compress=yes
#ProcessSizeMax=2G
#ExternalSizeMax=2G
#JournalSizeMax=767M`
```
And ExternalSize refers to uncompressed size.

smithfarm: Yes, and the default value is too low - I was seeing truncated coredumps.

smithfarm: > is the start here necessary?

It was not starting for me. I had to either (a) reboot or (b) systemctl start systemd-coredump.socket

Since rebooting in a vagrant provisioner is problematic (impossible?), I chose (b).

denisok: maybe something like `systemctl daemon-reload` is needed, so it is aware about that new socket.
But (b) is also ok

ddiss (approved):
lgtm

seslib/templates/provision.sh.j2
smithfarm: This does not seem to be necessary?

smithfarm: Ah, on SLE-15-SP2 it is not necessary. But in openSUSE the default is `| /bin/false`. I'll add it.

Thanks!

denisok (approved):

smithfarm:
```
    master: WWWW: core_dump_test
    master: Asserting that there are no existing coredumps in the system.
    master: + coredumpctl list
    master: + grep 'No coredumps found'
    master: No coredumps found.
    master: No coredumps yet. Good.
    master: Looking for osd.0 on this node
    master: + set +x
    master: ++ xargs
    master: ++ grep osd.0
    master: ++ pgrep -a ceph-osd
    master: Found "pgrep" line corresponding to osd.0:
    master: 45575 /usr/bin/ceph-osd -n osd.0 -f --setuser ceph --setgroup ceph --default-log-to-file=false --default-log-to-stderr=true --default-log-stderr-prefix=debug
    master: Extracted PID of osd.0: 45575
    master: Sending SIGSEGV to process 45575
    master: + osd_0_ps='45575 /usr/bin/ceph-osd -n osd.0 -f --setuser ceph --setgroup ceph --default-log-to-file=false --default-log-to-stderr=true --default-log-stderr-prefix=debug'
    master: + set +x
    master: + kill -SEGV 45575
    master: + sleep 30
    master: + coredumpctl list
    master: TIME                            PID   UID   GID SIG COREFILE  EXE
    master: Thu 2020-12-03 16:26:19 CET   45575   167   167  11 present   /usr/bin/ceph-osd
    master: + set +x
    master: core_dump_test: OK
```

smithfarm:
Thanks, everyone

Issue #567, opened 2020-12-03
caasp4 deployment broken


smithfarm:
`sesdev create caasp4` is currently broken. The spot where it fails is:

```
    master: ++ helm install --namespace metallb-system --name metallb stable/metallb -f /tmp/metallb.yaml
    master: Error: failed to download "stable/metallb" (hint: running `helm repo update` may help)
```

colder-is-better:
The culprit of that particular failure is a "rogue" line in `/etc/hosts` on the Master node (and on any of the Worker nodes):

```
127.0.1.1 master.mycaasp4.test master
```

Such a line prevents `skuba` from properly bootstrapping the cluster. One workaround is to have that line "weeded out" early on, e.g., like this:

```
diff --git a/seslib/templates/provision.sh.j2 b/seslib/templates/provision.sh.j2
index 25251c7..515e418 100644
--- a/seslib/templates/provision.sh.j2
+++ b/seslib/templates/provision.sh.j2
@@ -21,6 +21,7 @@ ls -lR /home/vagrant
 # populate /etc/hosts
 {% for _node in nodes %}
 {% if _node.public_address %}
+sed -ie "/127.0.1.1/d" /etc/hosts
 echo "{{ _node.public_address }} {{ _node.fqdn }} {{ _node.name }}" >> /etc/hosts
 {% endif %}
 {% endfor %}
```

I have tried this workaround and it works, and I am now investigating for the root cause of this issue.

smithfarm:
It's probably vagrant itself that is introducing this line. We used to remove it, but doing so breaks sesdev for folks who use the `--libvirt-networks` option, so we stopped removing it in https://github.com/SUSE/sesdev/pull/199

colder-is-better:
Yup, I also believe this is Vagrant's doing :/ How about we weed out that "rogue" line **only** when deploying a CaaSP cluster?

```
diff --git a/seslib/templates/caasp/master.sh.j2 b/seslib/templates/caasp/master.sh.j2
index c3f2b7a..2790978 100644
--- a/seslib/templates/caasp/master.sh.j2
+++ b/seslib/templates/caasp/master.sh.j2
@@ -3,6 +3,8 @@ zypper --non-interactive install --type pattern SUSE-CaaSP-Management
 
 {% if node.name == 'master' %}
 
+sed -ie "/127.0.1.1/d" /etc/hosts
+
 function wait_for_master_ready {
     set +ex
     echo "Waiting for master to be ready"
```

[...]

PR #568, opened 2020-12-05
Prevent Vagrant from setting the hostname in /etc/hosts

colder-is-better:
Having the hostname pointing to a local address can be
problematic, as was the case with skuba failing to
properly bootstrap the Master node.

See https://github.com/SUSE/sesdev/issues/567 for more.

### DIFF
--- a/seslib/templates/engine/libvirt/vagrant.node.j2
+++ b/seslib/templates/engine/libvirt/vagrant.node.j2
@@ -1,7 +1,5 @@
     {{ node.networks }}
 
-    node.vm.hostname = "{{ node.fqdn }}"
-
     node.vm.provider "libvirt" do |lv|
       lv.memory = {{ node.ram }}
       lv.cpus = {{ node.cpus }}


smithfarm:
My working assumption is that this breaks `--libvirt-networks`.

smithfarm:
@colder-is-better Could you please re-do the commit message so:

1. It has a `Signed-off-by:` line at the end
2. The last line of the description says `Fixes: https://github.com/SUSE/sesdev/issues/567` instead of "See #567 for more."

Thanks!

smithfarm:
@colder-is-better I can confirm that this patch causes Vagrant to stop adding the line to `/etc/hosts`, but it does not fix the issue #567 - that behavior remains the same.

Here is the test I did:

1. `sesdev --verbose create caasp4 --non-interactive --single-node --deploy-ses`  - ends with:

```
    master: ++ helm install --namespace metallb-system --name metallb stable/metallb -f /tmp/metallb.yaml
    master: Error: failed to download "stable/metallb" (hint: running `helm repo update` may help)
    master: +++ err_report 468
```

2. `sesdev ssh caasp4-mini`

3. `cat /etc/hosts`

```
master:~ # cat /etc/hosts
#
# hosts         This file describes a number of hostname-to-address
#               mappings for the TCP/IP subsystem.  It is mostly
#               used at boot time, when no name servers are running.
#               On small systems, this file can be used instead of a
#               "named" name server.
# Syntax:
#    
# IP-Address  Full-Qualified-Hostname  Short-Hostname
#

127.0.0.1       localhost

# special IPv6 addresses
::1             localhost ipv6-localhost ipv6-loopback

fe00::0         ipv6-localnet

ff00::0         ipv6-mcastprefix
ff02::1         ipv6-allnodes
ff02::2         ipv6-allrouters
ff02::3         ipv6-allhosts

10.20.120.200 master.caasp4-mini.test master
master:~ #
```

smithfarm:
This PR seems to have gone stale, so I'll close it. Feel free to reopen if more work on it is planned.

Issue #575, opened 2021-02-10
SES6 (devel) deployment broken


p-se:
The command
```bash
sesdev create ses6
```
leads to
```bash
    master: ++ zypper --non-interactive install vim git-core iputils jq make iptables patch man command-not-found bc apparmor-utils apparmor-parser chrony hostname systemd-coredump
node2: Loading repository data...
    node2: Reading installed packages...
    node2: 'vim' is already installed.
    node2: No update candidate for 'vim-8.0.1568-5.11.1.x86_64'. The highest available version is already installed.
    node2: 'command-not-found' is already installed.
    node2: No update candidate for 'command-not-found-0.2.2+20190613.e6c2668-6.3.2.noarch'. The highest available version is already installed.
    node2: 'hostname' is already installed.
    node2: No update candidate for 'hostname-3.16-2.22.x86_64'. The highest available version is already installed.
    node2: 'iputils' is already installed.
    node2: No update candidate for 'iputils-s20161105-8.3.1.x86_64'. The highest available version is already installed.
    node2: Resolving package dependencies...
    node2: Problem: systemd-coredump-234-24.64.1.x86_64 requires systemd = 234-24.64.1, but this requirement cannot be provided
    node2:   not installable providers: systemd-234-24.64.1.x86_64[update]
    node2:  Solution 1: Following actions will be done:
    node2:   downgrade of systemd-234-24.72.1.x86_64 to systemd-234-24.64.1.x86_64
    node2:   downgrade of systemd-sysvinit-234-24.72.1.x86_64 to systemd-sysvinit-234-24.64.1.x86_64
    node2:   downgrade of udev-234-24.72.1.x86_64 to udev-234-24.64.1.x86_64
    node2:  Solution 2: do not install systemd-coredump-234-24.64.1.x86_64
    node2:  Solution 3: break systemd-coredump-234-24.64.1.x86_64 by ignoring some of its dependencies
    node2: 
    node2: Choose from above solutions by number or cancel [1/2/3/c/d/?] (c): c
    node2: Error in sesdev deployment script trapped!
    node2: +++ err_report 203
    node2: +++ set +x
    node2: => hostname: node2
    node2: => script: /tmp/vagrant-shell
    node2: => line number: 203
    node2: Bailing out!
    master: Loading repository data...
    master: Reading installed packages...
    master: 'vim' is already installed.
    master: No update candidate for 'vim-8.0.1568-5.11.1.x86_64'. The highest available version is already installed.
[...]

PR #577, opened 2021-02-10
zypper: be more careful when installing systemd-coredump

smithfarm:
provision.sh: include zypper repo setup in --no-provision
    
    Recently a --no-provision feature was added to provide a way to quickly
    bring up a VM. But in order for that feature to be useful, the VM should
    have the same repos as in a full deployment.
    
---

    zypper: be more careful when installing systemd-coredump
    
    Due to the way Kiwi builds SLE Vagrant Boxes, the versions of packages
    pre-installed in the Vagrant Box may be newer than the versions of the same
    packages in the "channels" (i.e., in the zypper repos that are added after the
    VMs start).
    
    Work around this by force-installing systemd, udev, etc. from the repos.
    
    This has the drawback of putting the system into a state where "zypper ps -s"
    reports that it needs a reboot.
    
    Fixes: https://github.com/SUSE/sesdev/issues/575

---

    zypper: add "lsof" to basic set of packages to install
    
    The "lsof" package is required for "zypper ps -s" to work.
    Apparently, it is not pre-installed in at least one of the Vagrant Boxes that
    we use.
    
---

    provision.sh: inform user if reboot is needed
    
    Run "zypper ps -s" at the very end of the provisioning script.
    
---

    qa: add assert_reboot_not_needed test
    
    As part of qa-test, run "zypper ps -s" on all nodes and evaluate the output.
    
    If any node is in need of reboot, abort the test and instruct the user that
    it would be better if the cluster were rebooted before running qa-test.
    
    Since rebooting the cluster before running qa-test would have to be done
    manually, failure of this test currently only results in a "soft failure"
    (warning), not a hard failure.

---

Fixes: https://github.com/SUSE/sesdev/issues/575

### DIFF
--- a/qa/common/common.sh
+++ b/qa/common/common.sh
@@ -101,6 +101,57 @@ function ceph_log_grep_enoent_eaccess {
 # core validation tests
 #
 
+function assert_reboot_not_needed {
+    echo
+    echo "WWWW: assert_reboot_not_needed"
+    local success
+    success="yes"
+    local nodes_arr
+    local node_count
+    node_count="0"
+    local node_under_test
+    local n
+    local IFS
+    IFS=","
+    read -r -a nodes_arr <<<"$NODE_LIST"
+    for (( n=0; n<${#nodes_arr[*]}; n++ )) ; do
+        node_under_test="${nodes_arr[n]}"
+        node_under_test="${node_under_test//[$'\t\r\n']}"
+        node_count="$((node_count + 1))"
+        if ssh "$node_under_test" zypper ps -s | grep 'No processes using deleted files found' ; then
+            success="yes"
+        else
+            set +e -x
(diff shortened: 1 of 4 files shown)


smithfarm:
@p-se In my environment, this PR fixes #575 at least for `sesdev create ses6 --single-node`. What do you think?

p-se (approved):
works (even for `sesdev create ses6` (4 node) cluster)

smithfarm:
My only reservation is that, since this introduces the possibility that system processes will end up using deleted files, maybe the user should be made aware of this via some warning message.

smithfarm:
`contrib/standalone.sh` is a little rough these days (due to a small collection of issues unrelated to this PR), but I'm satisfied that this PR is now working as expected.

PR #579, opened 2021-03-12
cephadm: add 'period update' in radosgw deployment

trociny:
This command was added to the upstream doc in this PR [1],
and it fixes the issue with creating a rgw bucket via dashboard
(InvalidLocationConstraint error).

[1] https://github.com/ceph/ceph/pull/36496

Signed-off-by: Mykola Golub <mgolub@suse.com>

### DIFF
--- a/seslib/templates/cephadm/deployment_day_2.sh.j2
+++ b/seslib/templates/cephadm/deployment_day_2.sh.j2
@@ -230,6 +230,7 @@ touch {{ service_spec_gw }}
 radosgw-admin realm create --rgw-realm=default --default
 radosgw-admin zonegroup create --rgw-zonegroup=default --master --default
 radosgw-admin zone create --rgw-zonegroup=default --rgw-zone=default --master --default
+radosgw-admin period update --rgw-realm=default --commit
 cat >> {{ service_spec_gw }} << 'EOF'
 ---
 service_type: rgw


smithfarm:
@trociny This seems to break `sesdev --verbose create octopus --non-interactive --single-node octopus-1node`...

smithfarm:
Yup, without this commit the deployment succeeds.

trociny:
@smithfarm It is strange. I just tried `sesdev --verbose create octopus --non-interactive --single-node` and it worked for me.

What the error do you get? What does happen if you run `radosgw-admin period update --rgw-realm=default --commit` manually on a successfully deployed (not patched) cluster?

In my case I noticed the command requires some time to complete, so it might be some kind of timeout you are experiencing in your case?

smithfarm:
@trociny Sorry, I was confused. To see the failure you need to run `sesdev qa-test octopus-mini`. The failure is that the command

    sesdev create octopus --single-node

does not successfully deploy any rgw. It looks like cephadm tries to deploy rgw, but does not succeed:

```
master:~ # ceph orch ls --service-type rgw
NAME                 RUNNING  REFRESHED  AGE  PLACEMENT  IMAGE NAME  IMAGE ID   
rgw.default.default      0/1  -          -    master     <unknown>   <unknown>  
```

HOWEVER, the failure is not related to your PR. It happens on sesdev master as well. 

So, I began to suspect 15.2.9 and so far my testing indicates the problem is related to 15.2.9, and not to any change in sesdev. You can compare the following two:

    sesdev create ses7 --single-node  
    sesdev create ses7 --single-node --product

The former takes 15.2.9 from Devel:Storage:7.0, while the latter takes 15.2.8 (version currently shipping in the product). The former has broken RGW deployment and the latter does not.

What is strange, though, is that I ran this test on 15.2.9 when I put it in Devel:Storage:7.0 and it did not show any such problem.

smithfarm:
Annoyingly, I can no longer reproduce the rgw issue described in the previous comment.

smithfarm (approved):
@trociny Sorry this took so long to process. 

The issue I ran into when testing this PR was unrelated. For completeness, I reported the issue as https://bugzilla.opensuse.org/show_bug.cgi?id=1183740 and implemented a workaround in sesdev: #587

tl;dr: we recently dropped nfs-ganesha from the ceph-base container image that is built in the OBS. As a result, the "octopus" and "pacific" sesdev targets are no longer able to deploy clusters with the "nfs" role. When the nfs role is present, cephadm tries to deploy rgw-nfs and this causes the whole rgw to fail.

PR #581, opened 2021-03-12
sesdev: replace fancy list comprehension with simpleton code

smithfarm:
Though the replacement code looks like it was written by a simpleton (which it
was), it has two advantages over the code it is replacing:

1. it's easier to understand
2. it passes pylint

Fixes: https://github.com/SUSE/sesdev/issues/580

### DIFF
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -610,18 +610,25 @@ def _gen_settings_dict(
     for folder in synced_folder:
         try:
             src, dst = folder.split(':')
-            if not all([path.isabs(x) for x in [src, dst]]):
-                raise OptionValueError('--synced-folder',
-                                       "Please provide absolute paths for "
-                                       "synced folder paths",
-                                       folder)
-            if not path.exists(src):
-                raise OptionValueError('--synced-folder',
-                                       "Path to the source synced folder must exist",
-                                       src)
-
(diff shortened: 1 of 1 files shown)


sesdev/__init__.py
```diff
@@ -610,18 +610,25 @@ def _gen_settings_dict(
...
+                    folder
+                    )
+            if path.exists(src):
+                pass
+            else:
+                raise OptionValueError(
+                    "--synced-folder",
+                    "Path to the source synced folder must exist",
+                    src
+                    )
```
p-se, line 628: Just gave it a second thought and removing the square brackets makes it a generator already:

```suggestion
            if not all(path.isabs(x) for x in [src, dst]):
                raise OptionValueError('--synced-folder',
                                       "Please provide absolute paths for "
                                       "synced folder paths",
                                       folder)
            if not path.exists(src):
                raise OptionValueError('--synced-folder',
                                       "Path to the source synced folder must exist",
                                       src)
```

To be frank, I'd prefer it more concise, but it works in both cases.

smithfarm: I read somewhere that once Python programmers learn how to use comprehensions, they tend to use them everywhere. Maybe there was a **little** bit of that going on here...

PR #597, opened 2021-04-27
Custom provision config

p-se:
I've been using this for a while now and think it might also be useful for others. Although, it would also be useful for me to have it merged. It's not a full featured solution, it just enables to comfortably and automatically copy files to `root` and run an arbitrary script on each host on deployment. The custom provisioning can also be triggered after deployment for all VMs of a deployment or just one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ The Jenkins CI tests that `sesdev` can be used to deploy a single-node Ceph
       * [Run "make check" on openSUSE Leap 15.2 from upstream "octopus" branch](#run-make-check-on-opensuse-leap-152-from-upstream-octopus-branch)
       * [Run "make check" on SLE-15-SP2 from downstream "ses7" branch](#run-make-check-on-sle-15-sp2-from-downstream-ses7-branch)
       * [Other "make check" scenarios](#other-make-check-scenarios)
+   * [Custom provisioning](#custom-provisioning)
 * [Common pitfalls](#common-pitfalls)
    * [Domain about to create is already taken](#domain-about-to-create-is-already-taken)
    * [Storage pool not found: no storage pool with matching name 'default'](#storage-pool-not-found-no-storage-pool-with-matching-name-default)
@@ -1054,6 +1055,40 @@ $ sesdev create makecheck --os sles-15-sp2 \
 More combinations are supported than are described here. Compiling
 the respective `sesdev create makecheck` commands for these environments is left
 as an exercise for the reader.
+
+### Custom provisioning
+
+If you like to add configuration files or run arbitrary commands on each VM on
+deployment, you can do so by providing these files in the
+`~/.sesdev/.user_provision` directory.
+
+Note that all configuration files are copied to all the VMs on deployment, as
+well as the `provision.sh` file is executed on all VMs on deployment.
+
+#### Configuration files
+
+To have configuration added automatically to each VM, simply put them into the
+`~/.sesdev/.user_provision/config` directory. All files in this directory will be
+copied to `/root` on the hosts.
+
+##### Example
+
+Create a file `~/.sesdev/.user_provision/config/.vimrc` and it will be copied to
+`/root/.vimrc` on each host on deployment, so you will always have your personal
+Vim configuration on all hosts across all deployments.
+
+#### Running arbitrary commands
+
+Running any commands is achieved by creating a
+`~/.sesdev/.user_provision/provision.sh` file. The script will be executed after
+the deployment of a VM has been successfully completed.
+
+#### Triggering custom provisioning manually
+
+Custom provisioning can be triggered manually by issuing `sesdev user-provision
+<deployment-id>` or `sesdev user-provision <deployment-id> <host>` respectively.
+The former command applies the custom provisioning to all VMs in the deployment,
+whereas the latter variant only to a single VM.
 
 
 ## Common pitfalls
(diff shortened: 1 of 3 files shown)


seslib/deployment.py
```diff
@@ -349,10 +354,8 @@ def __generate_nodes(self):
             if self.settings.version == 'caasp4':
                 single_node = self.settings.single_node or len(self.settings.roles) == 1
                 if 'master' in node_roles or 'worker' in node_roles:
-                    if node.cpus < 2:
-                        node.cpus = 2
-                    if self.settings.ram < 2:
-                        node.ram = 2 * 2**10
+                    node.cpus = max(node.cpus, 2)
+                    node.ram = max(2 * 2**10, node.ram)
```
p-se, line 358: This was suggested by our linting, not sure if I have a more updated version as I'm also unable to explain where it comes from. It seems to have appeared of the void.

It was suggested for `cpus` only, but I also implemented it for `ram`. I'm not sure if you'll like it. Personally, I do. 

```
seslib/deployment.py:357:20: R1731: Consider using 'cpus = max(cpus, 2)' instead of unnecessary if block (consider-using-max-builtin)
seslib/deployment.py:390:16: R1731: Consider using 'cpus = max(cpus, 4)' instead of unnecessary if block (consider-using-max-builtin)
```

smithfarm: @p-se Either I don't understand your `node.ram` implementation, or it has lost the link to `self.settings.ram` being less than 2...

I'm fine with the `node.cpus` line since it's easy to understand.

I'm not fine with the `node.ram` line because I can't see how it's equivalent to the code it's replacing.

smithfarm:
@p-se Can you try moving the `pylint issues` commit into a separate branch and dropping it from this PR? I'd like to see the linter error that it is addressing.

smithfarm:
@p-se You wrote a very nice description of the first commit in the PR description:

```
I've been using this for a while now and think it might also be useful for others. Although, it would also be useful for me to have it merged. It's not a full featured solution, it just enables to comfortably and automatically copy files to root and run an arbitrary script on each host on deployment. The custom provisioning can also be triggered after deployment for all VMs of a deployment or just one.
```

Could you please add this text to the actual commit description as well?

README.md
```diff
@@ -1055,6 +1056,32 @@ More combinations are supported than are described here. Compiling
...
+`~/.sesdev/.provisioning` directory.
+
+Note that all configuration files are copied to all the VMs deployed, as well
+as the `provision.sh` file is executed on all VMs.
+
+#### Configuration files
+
+To have configuration added automatically to each VM, simply put them into the
+`~/.sesdev/.provisioning/config` directory. All files in this directory will be
+copied to `/root` on the hosts.
```
smithfarm, line 1072: Can you flesh this out a little by providing an example use case?

As a reader, I'm not sure what kind of configuration this would be adding...

seslib/deployment.py
```diff
@@ -1269,6 +1269,49 @@ def supportconfig(self, log_handler, name):
         ssh_cmd = ('rm', '/var/log/{}'.format(glob_to_get),)
         self.ssh(name, ssh_cmd)
 
+    def user_provision(self, node=None):
+        custom_provision_dir = os.path.join(Constant.A_WORKING_DIR, '.provision')
```
smithfarm, line 1276: this directory name `.provision` is confusing, because it seems like something internal to sesdev

Since the feature is called `user_provision`, it seems like the directory name itself should also be called `.user_provision`?

seslib/deployment.py
p-se: you're right, that's not how it's supposed to be. I'll fix it

p-se:
The issue fixed by the commit "pylint issues" has reappeared here, too: https://travis-ci.org/github/SUSE/sesdev/jobs/768553310

seslib/deployment.py
smithfarm: Could you please do that as a separate PR? We can merge that one first, and then the linter error will go away in this one as well.

smithfarm:
> The issue fixed by the commit "pylint issues" has reappeared here, too: https://travis-ci.org/github/SUSE/sesdev/jobs/768553310

Yes, thanks for letting me know. I'd prefer a minimalistic solution in a separate PR. (The code in question is no longer used or maintained, so there's no point in re-engineering it IMO.)

seslib/deployment.py
p-se: https://github.com/SUSE/sesdev/pull/598

README.md
p-se: Is that enough or would you say more about it?

p-se:
@smithfarm thank you very much for your review!

README.md
```diff
@@ -1055,6 +1056,38 @@ More combinations are supported than are described here. Compiling
...
+#### Running arbitrary commands
+
+Running any commands is achieved by creating a
+`~/.sesdev/.user_provision/provision.sh` file. The script will be executed after
+the deployment of a VM has been successfully completed.
+
+#### Triggering custom provisioning manually
+
+Custom provisioning can be triggered manually by issuing `sesdev user-provision
+<deployment-id>` or `sesdev user-provision <deployment-id> <host>` respectively.
```
smithfarm, line 1089: I like to say: "documentation is the art of stating the obvious". Could you please add here the following sentence:

"The former command applies the user provisioning to all hosts, whereas the latter variant only to a single host."

smithfarm: (I edited the proposed sentence several times - now I'm satisfied with it, I think!)

PR #598, opened 2021-04-27
fix pylint issues

p-se:

### DIFF
--- a/seslib/deployment.py
+++ b/seslib/deployment.py
@@ -349,8 +349,7 @@ class Deployment():  # use Deployment.create() to create a Deployment object
             if self.settings.version == 'caasp4':
                 single_node = self.settings.single_node or len(self.settings.roles) == 1
                 if 'master' in node_roles or 'worker' in node_roles:
-                    if node.cpus < 2:
-                        node.cpus = 2
+                    node.cpus = max(node.cpus, 2)
                     if self.settings.ram < 2:
                         node.ram = 2 * 2**10
                 if self.settings.caasp_deploy_ses or self.settings.explicit_num_disks:
@@ -382,8 +381,7 @@ class Deployment():  # use Deployment.create() to create a Deployment object
             if 'suma' in node_roles:
                 if self.settings.ram < 4:
                     node.ram = 4096
-                if self.settings.cpus < 4:
-                    node.cpus = 4
+                node.cpus = max(self.settings.cpus, 4)
                 # disk for /var/spacewalk
                 node.storage_disks.append(Disk(101))
                 # disk for /var/lib/pgsql


p-se:
@smithfarm I moved the changes to this new pull request as requested, but we can discuss here if it would make more sense to have those features removed, which, as you said, are not maintained nor used anymore.

Side note: Not relying on `self.settings.ram` is actually correct. A few lines above the adaptation, `node` is created from `self.settings` and hence `node.ram` contains the correct value already.

smithfarm:
I like the `cpu` fix: :+1: to that part. But the commit message is very terse. Could you add something like:

```
pylint started complaining as follows:

lint run-test: commands[1] | pylint seslib

************* Module seslib.deployment

seslib/deployment.py:357:20: R1731: Consider using 'cpus = max(cpus, 2)' instead of unnecessary if block (consider-using-max-builtin)

seslib/deployment.py:390:16: R1731: Consider using 'cpus = max(cpus, 4)' instead of unnecessary if block (consider-using-max-builtin)
```

smithfarm:
Further, I'd like to discuss dropping the `ram` part of this PR, for the following reasons:

1. pylint is not complaining about it, so it's not a "pylint issue" and there's no reason to fix it
2. as far as I know, this code is not (currently) used or maintained, so I don't understand why we should re-engineer it

smithfarm:
> @smithfarm I moved the changes to this new pull request as requested, but we can discuss here if it would make more sense to have those features removed, which, as you said, are not maintained nor used anymore.

I don't think we're ready to excise this code. It is not causing any harm and someone might want to revive it later.

p-se:
> Further, I'd like to discuss dropping the `ram` part of this PR, for the following reasons:
> 
> 1. pylint is not complaining about it, so it's not a "pylint issue"
> 2. as far as I know, this code is not (currently) used or maintained
> 
> In short, I don't see any reason to modify this code.

It is true that pylint does not complain about that part, however, taking the critique of it and thinking it further, it also applies to the `ram` part.

> Consider using '%s' instead of unnecessary if block Using the max builtin instead of a conditional improves readability and conciseness.

Using `max` we can safe an if block and improve readability as suggested by pylint. That what the idea I followed. If you like to only have fixed what is being complained about by pylint, I'm fine with that, too. I just wanted to have my motivation stated once. If you still prefer to only have fixed what was complained about, I'll just do that.

smithfarm:
Well, "improve readability" is subjective. Personally, I find the original code easier to understand.

I really would prefer a minimal solution here. This code is essentially frozen so I'm reluctant to change it any more than absolutely necessary.

p-se:
> Well, "improve readability" is subjective. Personally, I find the original `ram` code easier to understand.
> 
> I really would prefer a minimal solution here. This code is essentially frozen so I'm reluctant to change it any more than absolutely necessary.

So shall it be.

smithfarm:
Hopefully someone will eventually step forward to maintain these features (caasp4, suma). Otherwise, at some point the code should be ripped out so it doesn't cause irritations like these pylint errors.

Issue #601, opened 2021-07-30
Early provisioner failures trigger bogus hostname error message


smithfarm:
Because the provisioner (a bash script) uses `trap` and because the code triggered by `trap` runs `hostname`, the `hostname` binary must be installed as early as possible. 

That is not currently the case. The hostname binary currently gets installed only after repo setup has been completed. When the script errors out during repo setup, the error is trap and then the trap code produces another error when it tries to run `hostname` (which doesn't exist because the package is not installed).

PR #602, opened 2021-07-30
provision.sh: try harder to ensure hostname is installed

smithfarm:
77506e5e7211e998737b4ce0e058c148cbb771bf
deployment: run "vagrant up" with --no-destroy-on-error

Up to Vagrant 2.2.9, --no-destroy-on-error was the default and everything was
fine.

Then Vangrant 2.2.10 changed the default to --destroy-on-error. Since I was
ignorant of this option, I worked around it by merging
55314a49fb94c08a7681b72398eee33aaf9edef0 which ensured that the provisioner
script exits with status code 0 even when there is an error (that tricked
Vagrant into thinking the script succeeded, so it did not destroy the VMs when
the script failed).

Now that I know about this option, it makes sense to run with
--no-destroy-on-error since we don't want Vagrant to destroy VMs during
provisioning in any case (since it is assumed that we want to go in and debug to
determine why the provisioner failed).

---

47dc6592a9e112cbce490c761243a885d2bdc99d
provision.sh: try harder to ensure hostname is installed

When the provisioner fails before the hostname package is installed, we were
getting bogus error messages which sowed confusion.

Fixes: https://github.com/SUSE/sesdev/issues/601

### DIFF
--- a/seslib/deployment.py
+++ b/seslib/deployment.py
@@ -741,7 +741,7 @@ class Deployment():  # use Deployment.create() to create a Deployment object
             tools.run_async(cmd, log_handler)
 
     def _vagrant_up(self, node, log_handler):
-        cmd = ["vagrant", "up"]
+        cmd = ["vagrant", "up", "--no-destroy-on-error"]
         if self.existing:
             cmd.extend(["--no-provision"])
         else:
--- a/seslib/templates/provision.sh.j2
+++ b/seslib/templates/provision.sh.j2
@@ -1,19 +1,39 @@
-set -ex
+set -e
 
 DEPLOYMENT_SCRIPT="$0"
 
 function err_report {
+    local hn
     set +x
+    if hostname >/dev/null 2>&1 ; then
+        hn="$(hostname)"
+    else
+        hn="(cannot determine: hostname binary appears to be missing)"
+    fi
     local line_number="$1"
-    echo "Error in sesdev deployment script trapped!"
-    echo "=> hostname: $(hostname)"
+    echo "Error in provisioner script trapped!"
+    echo "=> hostname: $hn"
     echo "=> script: $DEPLOYMENT_SCRIPT"
     echo "=> line number: $line_number"
     echo "Bailing out!"
-    exit 0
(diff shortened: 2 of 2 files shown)


seslib/deployment.py
```diff
@@ -741,7 +741,7 @@ def _get_vagrant_box(self, log_handler):
             tools.run_async(cmd, log_handler)
 
     def _vagrant_up(self, node, log_handler):
-        cmd = ["vagrant", "up"]
+        cmd = ["vagrant", "up", "--no-destroy-on-error"]
```
p-se, line 744: Would it make sense to remove the changes added in 77506e5 at some point due to this option being added here?

smithfarm: I thought about reverting that commit, but decided against it. The commit actually adds value to the provisioner script. It's enough to change the exit value in `err_report` from 0 to 1 like I'm doing here.

seslib/templates/provision.sh.j2
```diff
@@ -1,19 +1,39 @@
...
     local line_number="$1"
-    echo "Error in sesdev deployment script trapped!"
-    echo "=> hostname: $(hostname)"
+    echo "Error in provisioner script trapped!"
+    echo "=> hostname: $hn"
     echo "=> script: $DEPLOYMENT_SCRIPT"
     echo "=> line number: $line_number"
     echo "Bailing out!"
-    exit 0
+    exit 1
```
smithfarm, line 19: @p-se This change from 0 to 1 effectively "reverts" https://github.com/SUSE/sesdev/commit/55314a49fb94c08a7681b72398eee33aaf9edef0 without losing any of the benefits that commit brought us. In other words, when the provisioner script fails, we will still get the useful error report. And `sesdev create` will go back to returning non-zero exit status on failure.

smithfarm:
@p-se I have tested this and it works as advertised.

seslib/deployment.py
smithfarm: Short answer: no. Long answer: see below.

PR #604, opened 2021-07-30
templates/zypper: install basic packages with --no-recommends

smithfarm:
When we install basic packages, we want just those packages and their hard
dependencies. We do not need or want any soft dependencies of those packages or
of their dependencies.

### DIFF
--- a/seslib/templates/zypper.j2
+++ b/seslib/templates/zypper.j2
@@ -217,9 +217,9 @@ zypper --non-interactive install --from update --force libncurses5 libncurses6
        'lsof',
    ] %}
 {% if os == 'sles-12-sp3' %}
-zypper --non-interactive install {{ basic_pkgs_to_install | join(' ') }} ntp
+zypper --non-interactive install --no-recommends {{ basic_pkgs_to_install | join(' ') }} ntp
 {% else %}
-zypper --non-interactive install {{ basic_pkgs_to_install | join(' ') }} chrony hostname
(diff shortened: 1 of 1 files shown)


smithfarm:
@p-se I don't think we need to be afraid to merge this, but I'm testing it anyway.

smithfarm:
NOTE: This part of the code, which installs some "basic packages" in the environment before deploying Ceph, is walking a fine line. On one hand, each additional package we install increases the risk that we will "hide" an undeclared dependency bug [1]. On the other hand, some packages are needed for sesdev to work properly, yet they cannot be classified as dependencies of any particular Ceph package. Or they *can* be classified as dependencies, but we need the packages to be installed as early as possible, to aid in debugging when the deployment fails.

In any case, not installing these packages with `--no-recommends` was a mistake, because without this option we end up installing a bunch of stuff (e.g., GUI stuff) that is completely unnecessary.

p-se:
@smithfarm 

Thanks for the explanation. I agree that it is unnecessary and possibly even masks issues if `--no-recommends` isn't used. I'll merge the PR once you've let me know that the tests have passed.

smithfarm:
TESTING:

- [x] ses5
- [ ] nautilus
- [ ] ses6
- [ ] octopus
- [ ] ses7

smithfarm:
@p-se The coredump test in octopus/ses7 is failing, but that regression is not related to this PR. I have opened https://bugzilla.suse.com/show_bug.cgi?id=1188979 to track it.

I think this PR is good to go.

PR #606, opened 2021-08-04
Fix --stop-before-cephadm-bootstrap

p-se:
`--stop-before-cephadm-boostrap` didn't work because the argument wasn't part of the code that reads the arguments and creates the context out of that. The context is then used to generate the templates (Vagrantfile and deployment scripts). This means that the switch never reached the conditions in the Jinja templates and was therefore technically ignored.

### DIFF
--- a/seslib/deployment.py
+++ b/seslib/deployment.py
@@ -610,6 +610,7 @@ class Deployment():  # use Deployment.create() to create a Deployment object
             'stop_before_ceph_salt_config': self.settings.stop_before_ceph_salt_config,
             'stop_before_ceph_salt_apply': self.settings.stop_before_ceph_salt_apply,
             'stop_before_ceph_orch_apply': self.settings.stop_before_ceph_orch_apply,
+            'stop_before_cephadm_bootstrap': self.settings.stop_before_cephadm_bootstrap,
             'reg': self.settings.reg,
             'image_path': self.settings.image_path,
             'use_salt': self.settings.use_salt,
--- a/seslib/settings.py
+++ b/seslib/settings.py
@@ -420,7 +420,7 @@ class Settings():
 
         config_tree = {}
 
-        def __fill_in_config_tree(config_param, global_param):
+        def _fill_in_config_tree(config_param, global_param):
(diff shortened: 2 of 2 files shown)


smithfarm:
```
Seemingly reappeared error in pylint. Thinking about renaming
`_fill_in_config_tree` to `fill_in_config_tree` to prevent this kind of
linting issues in the future, as renaming it to `__fill_in_config_tree`
was already a workaround for pylint when a single preceding underscore
failed.
```

Hm. It seems that disabling this linter check might be another option. If a check is obviously broken to the point that all it does is create noise and face-palming, then what is the use of running it?

seslib/deployment.py
```diff
@@ -610,6 +610,7 @@ def _generate_vagrantfile(self):
             'stop_before_ceph_salt_config': self.settings.stop_before_ceph_salt_config,
             'stop_before_ceph_salt_apply': self.settings.stop_before_ceph_salt_apply,
             'stop_before_ceph_orch_apply': self.settings.stop_before_ceph_orch_apply,
+            'stop_before_cephadm_bootstrap': self.settings.stop_before_cephadm_bootstrap,
```
smithfarm, line 613: Good catch :-)

smithfarm:
@p-se Are you interested in running `contrib/run-standalone.sh` on this PR, or should I do that?

(CAVEAT: It's time-consuming and sometimes uncovers unrelated problems that either have to be addressed or ignored.)

smithfarm:
Yay, the `continuous-integration/jenkins/pr-head` check is passing again :-)

(See https://bugzilla.suse.com/show_bug.cgi?id=1188979 for the ongoing saga of the bug that was causing it to fail.)

smithfarm (approved):

smithfarm:
TEST REPORT:

- [PASS] ses5
- [PASS] nautilus
- [PASS] ses6
- [PASS] octopus (octopus deployment is currently broken, but cause is not related to this PR)
- [ ] ses7
- [ ] pacific (currently broken)

PR #610, opened 2021-10-12
Switch to Github actions

p-se:


smithfarm:
It's strange though, because obviously the actions are running in this PR... sounds like we need more information.

p-se:
> @p-se ceph/ceph-salt is a completely different organization. Presumably they pay M$ for use of Actions.
> 
> It's possible that Actions will work for some time, before M$ starts demanding money and then they will stop working?

Quite possible and I do agree that we need more information. But also note that the repository you've referenced is private and that might be the game changing difference here. Anyway, I'll spend more time collecting information about Github Actions. Thanks for the hint!

smithfarm:
I'm going to merge this now: it seems like a clear improvement over the current state.

@p-se @kshtsk Please pursue the lint failure and other nits in a separate PR.

smithfarm:
oops - I failed to notice that this PR does not actually remove `.travis.yml`, so Travis CI is still active even after this PR merged.

So I'm definitively dropping Travis in #613 - jfyi

Issue #611, opened 2021-11-04
sesdev fails to deploy ses7 cluster due to changes in "ceph nfs cluster create" command


smithfarm:
When the "ceph nfs cluster create" command was first implemented, documentation was scarce. We got the command working by trial-and-error.

Recently, "sesdev create ses7" started to fail like this:

```
    master: ++ ceph nfs cluster create sesdev_nfs node3
    master: Error EINVAL: Invalid export type: sesdev_nfs
    master: +++ err_report 791
    master: +++ local hn
    master: +++ set +x
    master: Error in provisioner script trapped!
```

Nowaday's the syntax is:

```
$ ceph nfs cluster create <type> <clusterid> [<placement>]
```

where

```
<type> signifies the export type, which corresponds to the NFS Ganesha file system 
abstraction layer (FSAL). Permissible values are “cephfs” or “rgw”, but currently only
“cephfs” is supported.
```

Source: https://docs.ceph.com/en/octopus/cephfs/fs-nfs-exports/#create-nfs-ganesha-cluster

PR #612, opened 2021-11-04
templates/cephadm/deployment_day_2: fix NFS cluster create command

smithfarm:
At some point, upstream obviously changed the syntax of the "ceph nfs cluster
create" command in octopus, thereby breaking "sesdev create pacific" and "sesdev
create ses7".

Later, this syntax change got reverted [1], but since this revert was made
possible by a big feature [2], so far it has been backported only to pacific [3]

[1] https://github.com/ceph/ceph/pull/40411
[2] https://github.com/ceph/ceph/pull/37600
[3] https://github.com/ceph/ceph/pull/41005

Fixes: https://github.com/SUSE/sesdev/issues/611

### DIFF
--- a/seslib/templates/cephadm/deployment_day_2.sh.j2
+++ b/seslib/templates/cephadm/deployment_day_2.sh.j2
@@ -202,7 +202,11 @@ ceph fs status --format json-pretty
 
 {% if nfs_nodes > 0 %}
 {% set nfs_cluster = "sesdev_nfs" %}
+{% if version in ['octopus', 'ses7'] %}
+ceph nfs cluster create cephfs {{ nfs_cluster }} "$NFS_NODES_COMMA_SEPARATED_LIST"
+{% else %}
 ceph nfs cluster create {{ nfs_cluster }} "$NFS_NODES_COMMA_SEPARATED_LIST"
+{% endif %}{# if version in ['octopus', 'ses7'] #}
 sleep 10
 ceph nfs cluster ls
 {% endif %}{# nfs_nodes > 0 #}


trociny (commented):
LGTM

Just a note, in pacific the `type` (`cephfs`) argument has been removed again, so we will need to change this back (unless there is a way to make it work for both cases).

[1] https://github.com/ceph/ceph/pull/40411

smithfarm:
@trociny Thanks! My guess is that https://github.com/ceph/ceph/pull/37600 won't get backported to octopus, so https://github.com/ceph/ceph/pull/40411 won't get backported to octopus, either.

And yes there is a way to differentiate octopus from pacific here. I have updated the PR - please have another look?

seslib/templates/cephadm/deployment_day_2.sh.j2
```diff
@@ -202,7 +202,12 @@ ceph fs status --format json-pretty
 
 {% if nfs_nodes > 0 %}
 {% set nfs_cluster = "sesdev_nfs" %}
+{% if version in ['octopus', 'ses7'] %}
+ceph nfs cluster create cephfs {{ nfs_cluster }} "$NFS_NODES_COMMA_SEPARATED_LIST"
+{% endif %}{# if version in ['octopus', 'ses7'] #}
+{% if version in ['pacific', 'ses7p'] %}
```
mgfritch, line 208:
```suggestion
{% else %}
```
I encountered this and apparently created a similar fix in parallel (df948aee6351cc6e7ed08f6f66ce88a7904606a6), but I think we should simply put this in an `else` branch so that we don't need to revisit when Qunicy is released.

mgfritch (commented):
> @trociny Thanks! My guess is that [ceph/ceph#37600](https://github.com/ceph/ceph/pull/37600) won't get backported to octopus, so [ceph/ceph#40411](https://github.com/ceph/ceph/pull/40411) won't get backported to octopus, either.
> 

the plan is to not backport these PRs to Octopus and try to keep the CLI stable from Pacific forward ...

seslib/templates/cephadm/deployment_day_2.sh.j2
smithfarm: We have to revisit lots of stuff in sesdev for every Ceph major release, but hey, what's an if-else conditional among friends?

mgfritch (approved):

PR #614, opened 2021-11-05
sesdev: increase timeout in seconds

mgfritch:
avoid failures on less reasonable (slow) hardware

Signed-off-by: Michael Fritch <mfritch@suse.com>


mgfritch:
one of my machines is resource constrained and frequently encounters this timeout while installing packages using `zypper` ...

smithfarm (approved):

PR #617, opened 2021-11-17
pycodestyle: Fix issue with yaml.load (use safe_load)

p-se:
Replace `yaml.load` with `safe_load` to fix pycodestyle issue and ensure
using tests that `safe_load` is capable of loading the `config.yaml`
file correctly, including the ability to use the YAML reference (which
is not explicitly documented in the sesdev docs but part of YAML).

`yaml.load` has been to be used without a Loader. However it is unsafe and
has been used way too frequently, although the documentation of PyYAML
has always stated that it is unsafe.

Fixes: #618 

### DIFF
--- a/tests/test_parse_config_yaml.py
+++ b/tests/test_parse_config_yaml.py
@@ -1,0 +1,52 @@
+from seslib.settings import Settings
+from seslib.constant import Constant
+
+
+from unittest.mock import patch, mock_open
+from tempfile import mktemp
+
+
+def test_parse_config_yaml__return_empty_dict():
+    Constant.CONFIG_FILE = mktemp()
+    config_tree = Settings._load_config_file()
+    assert config_tree == {}, 'reading non-existent file does not return empty dict'
+
+
+# Examples from the documentation of how config.yaml can be used.
+# Excluding the ability to use references in YAML, that is not from the docs but made up.
+config_yaml = '''
+libvirt_use_ssh: true
+libvirt_user: root
+libvirt_private_key_file: $HOME/.ssh/id_rsa
+libvirt_host: foo
+version_devel_repos:
+  octopus:
+      leap-15.2:
+          - 'https://download.opensuse.org/repositories/filesystems:/ceph:/octopus/openSUSE_Leap_15.2'
+image_paths_devel:
+    octopus: 'registry.opensuse.org/filesystems/ceph/octopus/images/ceph/ceph'
+container_registry:
+    prefix: 'registry.suse.de'
+    location: '1.2.3.4:5000'
+    insecure: True
+version_default_roles: &default
+    octopus:
+        - [master, mon, mgr, storage]
+        - [mon, mgr, storage]
+        - [mon, mgr, storage]
+other_roles: *default
+    '''  # noqa
+
+
+@patch('os.path.exists', return_value=True)
+@patch('os.path.isfile', return_value=True)
+@patch('builtins.open', mock_open(read_data=config_yaml))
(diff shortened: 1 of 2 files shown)


smithfarm:
@p-se LGTM! The only nit I have is: it says it is fixing a bug, but the bug apparently isn't reported anywhere.

Could you open an issue using https://github.com/SUSE/sesdev/issues/new and add an appropriate `Fixes:` line to the commit message and PR description?

tests/test_parse_config_yaml.py
```diff
@@ -0,0 +1,52 @@
...
+    octopus:
+        - [master, mon, mgr, storage]
+        - [mon, mgr, storage]
+        - [mon, mgr, storage]
+other_roles: *default
+    '''  # noqa
+
+
+@patch('os.path.exists', side_effect=lambda _: True)
+@patch('os.path.isfile', side_effect=lambda _: True)
```
kshtsk, line 42: just curious, why the lambda is required here?

kshtsk:
> @p-se LGTM! The only nit I have is: it says it is fixing a pycodestyle-related bug, but this bug (apparently?) was never reported and is not being tracked?
> 
> Could you open an issue using https://github.com/SUSE/sesdev/issues/new and add an appropriate `Fixes:` line to the commit message and PR description?

This is overkill. It is not even a bug, but linter error message.

smithfarm (approved):

smithfarm:
> It is not even a bug, but linter error message.

Oh. I thought `yaml.load` was unsafe (an unreported security issue, not just a linter issue). I guess I was wrong!

p-se:
> > It is not even a bug, but linter error message.
> 
> Oh. I thought `yaml.load` was unsafe (an unreported security issue, not just a linter issue). I guess I was wrong!

Well, not in the sense that there's an unreported CVE, but in the same sense that pickle is insecure and should never be used on untrusted sources. Arbitrary commands can be executed. In the past, and although `load` was never intended to be safe, even a CVE was created because of that. I'd guess the design decision could have been better; making `load` secure and providing `unsafe_load` or similar for all other purposes. Seemingly the capabilities of `safe_load`, which are limited, are still very sufficient for most use cases (as in our case). It shouldn't harm to use a safe function by default.

tests/test_parse_config_yaml.py
p-se: Let me check. It might turn out not to be. But in general, the lambda creates a replacement function for `os.path.exists` and `os.path.isfile`. Both are called with a single argument and, in the case of the test, are supposed to simply return `True`. But maybe `return_value` does the trick, too.

p-se: It obviously was **not** necessary, as `return_value` worked similarly well. So, I switched to that.

smithfarm:
This is a total nit, but could you change the fixes line so it contains the full URL of the issue? I.e.

```
Fixes: https://github.com/SUSE/sesdev/issues/618
```

(`Fixes: #618` is ambiguous and requires the reader to have special knowledge - and do additional work - if they want to find and display the issue.)

p-se:
> This is a total nit, but could you change the fixes line so it contains the full URL of the issue? I.e.
> 
> ```
> Fixes: https://github.com/SUSE/sesdev/issues/618
> ```
> 
> (`Fixes: #618` is ambiguous and requires the reader to have special knowledge - and do additional work - if they want to find and display the issue.)

I'll make it a habit.

kshtsk (approved):

PR #621, opened 2021-12-03
provision: increase max number loops (50 -> 500)

mgfritch:
avoid `Looped too many times` error on resource constrained hardware

Signed-off-by: Michael Fritch <mfritch@suse.com>


ml8mr (approved):
LGTM (since this is required on our super fast test hardware).

p-se (approved):

PR #631, opened 2022-02-28
tox: speed up linting

p-se:
From previously ~15s to ~7s by utilizing more cores using the `-j0` command line argument of `pylint`.

Note: I'm using `tox -p` to start the two tox jobs in parallel, one of which is being sped up by this PR.


mgfritch (approved):
seems reasonable to me

PR #632, opened 2022-03-01
Deployment: Set flag to do not show the 'Expand cluster' page

votdev:
… in the Ceph Dashboard after signing in the first time.

Signed-off-by: Volker Theile <vtheile@suse.com>


p-se:
Doesn't the upstream implementation fix that issue? https://github.com/ceph/ceph/pull/44951/files

votdev:
> Doesn't the upstream implementation fix that issue? [ceph/ceph#44951 (files)](https://github.com/ceph/ceph/pull/44951/files)

No, it only takes care to set the flag the cluster is NOT installed.

votdev:
To avoid confusion the PR has been closed as it may be hiding something that may occur in a customer installation.

Issue #633, opened 2022-03-03
ERROR: SUSE dashboard branding appears to be completely absent!


p-se:
An issue in Podman 2.2.1 breaks the QA test by adding a line containing the IP `127.0.1.1` to the `/etc/hosts` file inside the containers, when they are being spawned using `--network=host`. This only affects the `octopus` deployment, as `ses7` has an older Podman version 2.1.1 where this problem does not occurr.

https://github.com/containers/podman/issues/10319

When the container issues `socket.getfqdn()` the name of the container (e.g. `ceph-ba72a5f2-9a0c-11ec-a714-5254000f2321-mgr.master.yhzbor`) is being returned, thereby breaking what Ceph reports in `ceph mgr services`. As that name is not accessible from the host VM or other nodes, the QA test breaks.

`sesdev qa-test <octopus-deployment>` breaks and reports

```
   [...]
    master: All repositories have been refreshed.
    master: + _zypper_install_on_master curl
    master: + set -x
    master: + zypper --non-interactive install --no-recommends curl
    master: Loading repository data...
    master: Warning: Repository 'Main Update Repository' appears to be outdated. Consider using a different mirror or server.
    master: Reading installed packages...
    master: Resolving package dependencies...
    master: 
    master: The following NEW package is going to be installed:
    master:   curl
    master: 
    master: 1 new package to install.
    master: Overall download size: 331.8 KiB. Already cached: 0 B. After the operation, additional 471.2 KiB will be used.
    master: Continue? [y/n/v/...? shows all options] (y): y
    master: Retrieving package curl-7.66.0-lp152.3.24.1.x86_64 (1/1), 331.8 KiB (471.2 KiB unpacked)
    master: Retrieving: curl-7.66.0-lp152.3.24.1.x86_64.rpm [done]
    master: 
    master: Checking for file conflicts: [.....done]
    master: (1/1) Installing: curl-7.66.0-lp152.3.24.1.x86_64 [...........done]
    master: + set +x
    master: ++ ceph mgr services
    master: ++ jq -r .dashboard
    master: + dashboard_url=https://ceph-ba72a5f2-9a0c-11ec-a714-5254000f2321-mgr.master.yhzbor:8443/
    master: + grep -i suse
    master: + curl --silent --insecure https://ceph-ba72a5f2-9a0c-11ec-a714-5254000f2321-mgr.master.yhzbor:8443/
    master: ERROR: SUSE dashboard branding appears to be completely absent!
[...]

PR #634, opened 2022-03-03
seslib: Ignore Podman 2.2.1 in `octopus`

p-se:
Ignore Podman 2.2.1 in `octopus` as it contains an unfixed bug that, in
conjunction with using `--network=host` leads to an `/etc/hosts` entry
being added which confuses Pythons `socket.getfqdn()` function and leads
to inaccessible links in `ceph mgr services`.

Fixes: https://github.com/SUSE/sesdev/issues/633

### DIFF
--- a/seslib/templates/zypper.j2
+++ b/seslib/templates/zypper.j2
@@ -39,6 +39,11 @@ zypper --non-interactive removerepo repo-source-non-oss || true
 zypper addrepo --refresh {{ os_repo_url }} {{ os_repo_name }}
 {% endfor %}
 
+# Ignore Podman 2.2.1 due to bug: https://github.com/SUSE/sesdev/issues/633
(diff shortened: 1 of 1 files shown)


p-se:
😁

smithfarm:
LGTM but if I were doing this I would put this code in `zypper.j2` (the idea was not to have `zypper` commands scattered all through the Jinja code, but to concentrate them all in one place).

https://github.com/SUSE/sesdev/blob/master/seslib/templates/zypper.j2

kshtsk (approved):
Looks trivial, I didn't check it but if you sure ok to merge

p-se:
> Looks trivial, I didn't check it but if you sure ok to merge

The `continuous-integration/jenkins/pr-head` check finally went through because of this :)

PR #645, opened 2022-04-14
constants: ses7.1 repo urls

m-ildefons:
- Uncomment and rectify ses7.1 repo urls for sles-15-sp3


colder-is-better (approved):
Looks good to me.

Issue #652, opened 2022-05-25
feature request: deploy OSDs with mixed disk layouts


m-ildefons:
Currently, nodes with the `storage` role have a given number of disk drives, all with the same given size, but sometimes it is required that nodes are provisioned with a mix of drives with different characteristics, size being the easiest one to modify.
These extravagant disk layouts could be created by specifying `--num-disks` and `--disk-size` multiple times, e.g.
```bash
$ sesdev create ses7p --num-disks 2 --disk-size 10 --num-disks 2 --disk-size 20
```
resulting in a ses7p cluster where each node with the `storage` role has two drives of 10GiB size and two drives of 20 GiB size in addition to the root drive.
When `--disk-size` is not specified often enough, the remaining drives will be created with default size, if it is specified more often than `--num-disks`, the superfluous arguments will be ignored.
This feature would be a great help when testing OSD deployment processes and drive group configuration.

PR #653, opened 2022-05-25
feature: mixed disk drive layouts

m-ildefons:
Provision mixed disk drive layouts when specifying `--num-disks` and
`--disk-size` multiple times.

fixes: #652

### DIFF
--- a/seslib/deployment.py
+++ b/seslib/deployment.py
@@ -234,7 +234,7 @@ class Deployment():  # use Deployment.create() to create a Deployment object
         self.settings.override('single_node', True)
         self.settings.override('roles', Constant.ROLES_DEFAULT_BY_VERSION['makecheck'])
         if not self.settings.explicit_num_disks:
-            self.settings.override('num_disks', 0)
+            self.settings.override('num_disks', [0])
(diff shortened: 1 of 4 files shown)


seslib/deployment.py
```diff
@@ -469,6 +466,12 @@ def __generate_nodes(self):
 
             self.nodes[node.name] = node
 
+    def __append_disks_to_node(self, node):
+        for _i, _n in enumerate(self.settings.num_disks):
+            for _ in range(_n):
+                node.storage_disks.append(Disk(self.settings.disk_size[_i]
```
kshtsk, line 472: What happens disk_size length is less than `num_disks`?

kshtsk (changes requested):
I don't think it is ready to merge yet.

kshtsk:
jenkins retest this please

seslib/deployment.py
m-ildefons: I wasn't aware that the `foo[bar] or buzz` syntax doesn't work with lists. AFAIK it works with dicts though. Fixed it.

PR #654, opened 2022-06-23
ses6: set dashboard admin password via grain

tserong:
By default, DeepSea will set a random password for the dashboard admin user, by calling salt's random.get_str function:

https://github.com/SUSE/DeepSea/blob/64cf6c350da1214cc0bbcc11e372b25cbd29d35a/srv/salt/ceph/dashboard/default.sls#L2

It's possible for the string returned to include characters that break the parsing of the above jinja template and/or screw up the line that echoes the password into `ceph dashboard ac-user-create`.  We can work around this potential problem by setting the admin password to "admin" in advance of stage 3, using the dashboard_creds grain.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1200794


m-ildefons (approved):
Thanks Tim, looks good to me!

PR #656, opened 2022-07-07
feature: generate maintentance test reports

m-ildefons:
- Add command for generating maintenance test report data on a running
  cluster.

When testing maintenance updates for various packages, certain
informations need to be collected and entered into the final report.
Thus far, the information was largely collected by hand from various
commands run against the cluster.

This change adds a new sub command, which aggregates this information
and prints it ready for copy-and-paste to the console.
Included in this information is
 - The command and configuration used to create the test cluster
 - Basic information about the resulting cluster layout
 - A list of packages and installation statuses from each relevenat
   repo on each host
 - The Ceph cluster status

### DIFF
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -1324,6 +1324,53 @@ def show(deployment_id, **kwargs):
 
 @cli.command()
 @click.argument('deployment_id')
+def report(deployment_id):
+    """
+    Compile the relevant information for the maintenance test report. This
+    includes cluster-wide information, individual VM information and package
+    installation statuses.
+    """
+    def _indent(string, spaces=4):
+        return " " * spaces + string
+
+    dep = Deployment.load(deployment_id)
+    ver = Constant.VERSION_OFFICIAL[dep.settings.version]
+    cmd = f"sesdev create {dep.settings.version} --product --repo-priority"
+    for rep in dep.settings.custom_repos:
+        cmd += " --repo {}".format(rep['url'])
+    cmd += f" {deployment_id}"
+    dep.start(_print_log)
+
+    packages = dep.list_packages(repos=[r['url'] for r in dep.settings.custom_repos])
+
+    click.echo("<QAM-SES>")
+    click.echo("")
+    click.echo("* sesdev was used for installation/smoke testing")
+    click.echo("")
+    click.echo(f"* Command for creating a {ver} test cluster:")
+    click.echo(_indent(cmd))
+    click.echo("")
+    click.echo(f"* {ver} cluster characteristics:")
+    for line in dep.configuration_report(show_individual_vms=True).splitlines():
+        click.echo(_indent(line))
+    click.echo("")
+    click.echo("* Packages installed from Maintenance repos:")
+    for node, pkgs in packages.items():
+        click.echo("")
+        click.echo(f" -- {node}:")
+        for pkg in pkgs:
+            click.echo(_indent(f"{pkg.name}  {pkg.version}  (from {pkg.repo})"))
+    click.echo("")
+    click.echo(f"* {ver} cluster status:")
+    click.echo("")
+    dep.ssh('master', ['ceph', 'status'], interactive=False)
+    click.echo("")
+    click.echo("* Workflow QAM-SES: PASSED/FAILED")
+    click.echo("</QAM-SES>")
+
+
+@cli.command()
+@click.argument('deployment_id')
 @click.argument('node', required=False)
 @click.argument('command', required=False, nargs=-1, type=click.Path())
 def ssh(deployment_id, node=None, command=None):
--- a/seslib/deployment.py
+++ b/seslib/deployment.py
@@ -52,7 +52,7 @@ from .exceptions import \
 from .log import Log
 from .node import Node, NodeManager
 from .settings import Settings, SettingsEncoder
-from .zypper import ZypperRepo
+from .zypper import ZypperRepo, ZypperPackage
 
 
 class Disk():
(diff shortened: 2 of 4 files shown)


seslib/deployment.py
```diff
@@ -1815,6 +1815,16 @@ def replace_mgr_modules(self, local=None, pr=None, branch=None, repo=None, langs
             ssh_cmd.append("ceph mgr module enable {}".format(module))
             tools.run_sync(ssh_cmd)
 
+    def list_packages_from_repo(self, repo):
```
kshtsk, line 1818: For generality refactor the function to `list_packages` with optional _repo_ argument, which returns all packages for all repos?

kshtsk: Also, the function lacks of description, and I guess it should return a dictionary of nodes to list of package names rather than list of lines of output from zypper command execution.

kshtsk (changes requested):
See comments above.

sesdev/__init__.py
```diff
@@ -1322,6 +1322,56 @@ def show(deployment_id, **kwargs):
...
 
 
+@cli.command()
+@click.argument('deployment_id')
+def maintenance_report(deployment_id):
+    """
+    Compile the relevant information for the maintenance test report. This
+    includes cluster-wide information, individual VM information and package
+    installation statuses.
+    """
```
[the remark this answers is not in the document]
m-ildefons: `status` exists already. `report` may be a good idea, but this is report specialized to the needs of the QA maintenance workflow.

seslib/deployment.py
m-ildefons: I'd like to avoid constructing very generalized functions when they are not needed. Also I'd like to avoid parsing `zypper` output. I'll see if I can fetch the data in a machine-friendly format like JSON...

kshtsk: @m-ildefons but you're doing the parsing already in the `__init__` module, all I say it is to bring it here and make the code more functional.

sesdev/__init__.py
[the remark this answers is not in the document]
kshtsk: From what I see in the description of the function, there is nothing specific, what not be useful for the other deployment.

seslib/deployment.py
kshtsk: And by the way, the unit test for this function would be excellent accompaniment.

seslib/deployment.py
```diff
@@ -1815,13 +1815,40 @@ def replace_mgr_modules(self, local=None, pr=None, branch=None, repo=None, langs
...
         packages = {}
         for node in self.nodes:
             ssh_cmd = self._ssh_cmd(node)
-            ssh_cmd.append("zypper pa -N -R -i -r {}".format(repo))
-            plist = tools.run_sync(ssh_cmd)
-            packages[node] = plist
+            ssh_cmd.append(zypper_cmd)
+            raw_package_list = tools.run_sync(ssh_cmd)
+
+            packages[node] = _to_package_list(raw_package_list)
```
kshtsk, line 1851: Looks good, but maybe move this code to zypper module?

kshtsk: and a unit test for parsing package names from zypper output?

m-ildefons: I'd like to keep the code in the deployment module because it operates on a deployments private data.

I haven't added unit tests for one simple reason, the call to zypper and the parsing are tied together, because the parser function expects the zypper output in a format specific to this call. Therefore unless I am to fake out a ssh connection, there really is no good way to unit test this function and that is very hard to do for code that sees this little use. We don't need NASA level code quality here.

kshtsk: It is logical to place zipper output parsing code in zipper module, since it is universal. It is not about adding full coverage for NASA type code quality, just basic parsing thing that makes it simple to verify the code without deploying the stuff, the easiest way is to mock the ssh call or isolate the parsing code from the calls ssh.

m-ildefons: It is *not* universal. The parsing code assumes that the output of zypper is in a special non-default format. Therefore the parsing code depends on the details of the ssh call.
It is also not universal because other zypper subcommands will produce other formats, which it will not parse either.

All I want is a list of packages from the specified repos and their installation status/version and I want that in the report. I would even take the output of zypper verbatim (which was my first implementation of this). I don't want to write universal zypper output parsers and I don't want to write code that someone might re-use (which is highly unlikely given the state of SES). The zypper module is a stub and I'd like to keep it that way because writing python bindings for zypper is nice, but out of scope for this project.

kshtsk: > It is _not_ universal. The parsing code assumes that the output of zypper is in a special non-default format. Therefore the parsing code depends on the details of the ssh call. It is also not universal because other zypper subcommands will produce other formats, which it will not parse either.

From my point of view it is universal when it comes to get a list of packages on from any node where zypper is installed.
And thus there should be a test coverage for the code which is parsing the output for that set of options for zypper. Since it is parsing output from the zypper, it is natural to put it in zypper module, there is nothing specific to a deployment. The list_packages could be left in deployment but only as a proxy call to responding zypper, for example, static methods.

kshtsk:
And by the way when we done with addressing hints to the PR it is better to squash all the commits within it before merging.

seslib/deployment.py
m-ildefons: Read the code:
```python
self._ssh_cmd(node)
```
is not accessible outside of the context of a deployment. The functionality of `list_packages` is tightly coupled to deployments. And the functionality of `_to_package_list` is coupled to specific zypper options. It would be a bad idea to bend over backwards in order to decouple the code and be able to insert unit tests (which btw amount to nothing more than making sure pythons `str.split()` works.

You can make a case out of missing support for ubuntu based deployments, but then again that is irrelevant for what I intend with this PR.

kshtsk: > Read the code:
> 
> ```python
> self._ssh_cmd(node)
> ```
> 
> is not accessible outside of the context of a deployment. The functionality of `list_packages` is tightly coupled to deployments. And the functionality of `_to_package_list` is coupled to specific zypper options. It would be a bad idea to bend over backwards in order to decouple the code and be able to insert unit tests (which btw amount to nothing more than making sure pythons `str.split()` works.
> 
I can see what you're trying to say, but I don't think it is THAT bad idea. Alternatively you might use mocking, but decoupling looks easier to refactor the code. Although embedded functions is a great mechanism, but only for some simple piece of code, the bad thing is that it is impossible to test them, so more or less complicated unit should be decoupled. 
It is not only how split works, it is also, which indexes for table records and for which arguments do use it for ZypperPackage, also it is about different skipping code you've been using.

> You can make a case out of missing support for ubuntu based deployments, but then again that is irrelevant for what I intend with this PR.

No, we don't care about Ubuntu here at all now, and RedHat neither.

seslib/deployment.py
```diff
@@ -1815,6 +1815,45 @@ def replace_mgr_modules(self, local=None, pr=None, branch=None, repo=None, langs
             ssh_cmd.append("ceph mgr module enable {}".format(module))
             tools.run_sync(ssh_cmd)
 
+    def list_packages(self, repos=None):
```
kshtsk, line 1818: Still needs to be covered with unit tests.

asettle:
> And by the way when we done with addressing hints to the PR it is better to squash all the commits within it before merging.

This has been open for far too long. Let's squash, and merge. I appreciate that the unit tests are best practice, but sesdev can't be taking this much time from the both of you. 

We can revisit this and the need unit tests if the time arises.

kshtsk (approved):
ok for now

PR #657, opened 2022-07-08
feature: create test deployment from maintenance id

m-ildefons:
- Create a test deployment with auto-generated settings from a
  maintenance id

Maintenance update package tests rely on a consisted test environment to
produce reliable results. This change codifies the settings used to
generate such an environment, so that operator errors are minimized.
This convenience command takes a maintenance id of the form
'S:M:xxxxx:yyyyyy' and generates a cluster with the right additional
repos and settings for reliable testing.

### DIFF
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -4,6 +4,7 @@ import logging
 from os import environ, path
 import re
 import sys
+import requests
 
 from prettytable import PrettyTable
 
@@ -1666,3 +1667,116 @@ def shell_completion(shell):
     completion = tools.run_sync('sesdev')
     print(completion)
     environ['_SESDEV_COMPLETE'] = ''
+
+
+@cli.command()
+@click.argument('review_request_id')
+@click.argument('version')
+@common_create_options
+@deepsea_options
+@ceph_salt_options
+@libvirt_options
+@ipv6_options
+def maintenance_test(review_request_id, version, **kwargs):
+    """
+    Create a test cluster for installation/smoke testing of package maintenance
+    updates. Takes a review request id of the form 'SUSE:Maintenance:xxxxx:yyyyyy'
+    or 'S:M:xxxxx:yyyyyy' and a SES version and creates a test cluster.
+
+    The difference between the test cluster and a 'normal' sesdev cluster is
+    that the test cluster has exra repositories from which it installs the
+    packages under testing.
+
+    This function basically hard-codes
+
+        sesdev create {version} --product --repo-priority \
+          --repo http://download.suse.de/ibs/SUSE:/Maintenance:/xxxxx/SUSE_SLE-15_Update \
+          --repo \
+          http://download.suse.de/ibs/SUSE:/Maintenance:/xxxxx/SUSE_Updates_Storage_{vnum}_x86_64 \
+          {xxxxx}-{yyyyy}-{version}
+    """
+    def _check_url(url):
+        try:
+            get = requests.get(url)
+
+            if get.status_code == 200:
+                return True
+            return False
+        except requests.exceptions.RequestException:
+            return False
+
+    def _parse_review_request_id(rrid):
+        """
+        Takes a Review Request ID of the form S:M:xxxxx:yyyyyy or
+        SUSE:Maintenance:xxxxx:yyyyyy as a string and returns just the
+        maintenance incident id and the review reuqest number as a pair of
+        strings. If the given argument does not sufficiently conform to the
+        format of a Review Request ID, a ValueError with a descriptive message
+        is raised.
+
+        Example input:
+
+            "SUSE:Maintenance:12345:678901"
+
+        Example output:
+
+            ("12345", "678901")
+        """
+        try:
+            suse, maintenance, maintenance_incident, review_request = rrid.split(':')
+            if suse not in ['SUSE', 'S']:
+                raise ValueError("Not a Review Request ID. Must begin with `SUSE:` or `S:`")
+            if maintenance not in ['Maintenance', 'M']:
+                raise ValueError("Not a Review Request ID. Second part must be"
+                                 " `Maintenance:` or `M:`")
+            if 0 > int(maintenance_incident) or 100000 < int(maintenance_incident):
+                raise ValueError("Not a Review Request ID. The Maintenance"
+                                 " Incident ID must be a positive five digit number.")
+            if 0 > int(review_request) or 1000000 < int(review_request):
+                raise ValueError("Not a Review Request ID. The Review Request Number"
+                                 " must be a positive six digit number.")
+
+            if not _check_url(f"http://qam2.suse.de/reports/SUSE:Maintenance:\
+{maintenance_incident}:{review_request}"):
+                raise ValueError(f"Not a valid Review Request ID. The Review Request ID \
+'SUSE:Maintenance:{maintenance_incident}:{review_request}' \
+does not exist.")
+            return (maintenance_incident, review_request)
+        except ValueError as value_error:
+            click.echo(f"{value_error}")
+            click.echo(f"Review Request ID {rrid} does not conform to pattern S:M:xxxxx:yyyyyy")
+            sys.exit(1)
+
+    maintenance_incident_id, review_request_number = _parse_review_request_id(review_request_id)
+    deployment_id = f"{maintenance_incident_id}-{review_request_number}-{version}"
+
+    repo_urls = {}
+
+    for alias, url in Constant.MAINTENANCE_REPO_TEMPLATES[version].items():
+        if _check_url(url.format(maintenance_incident_id)):
+            repo_urls[alias.format(maintenance_incident_id)] = url.format(maintenance_incident_id)
+
+    if len(repo_urls) == 0:
+        click.echo('Could not find any maintenance repos.')
+        click.echo('Please make sure you are connected to the E&I VPN.')
+        click.echo('\nThe following URLs were probed:')
+        for alias, url in Constant.MAINTENANCE_REPO_TEMPLATES[version].items():
+            click.echo('  - ' + alias.format(maintenance_incident_id) + ': ' +
+                       url.format(maintenance_incident_id))
+        click.echo('Aborting...')
+        sys.exit(1)
+
+    _prep_kwargs(kwargs)
+    settings_dict = _gen_settings_dict(version, **kwargs)
+    settings_dict['devel_repo'] = False
+    settings_dict['repo_priority'] = True
+
+    for repo_name, repo_url in repo_urls.items():
+        settings_dict['custom_repos'].append(
+            {
+                'name': repo_name,
+                'url': _maybe_munge_repo_url(repo_url),
+                'priority': Constant.ZYPPER_PRIO_ELEVATED
+            })
+
+    _create_command(deployment_id, settings_dict)
--- a/seslib/constant.py
+++ b/seslib/constant.py
@@ -124,6 +124,35 @@ class Constant():
 
     LOGFILE = None
 
+    # A dict of string templates for repo urls. These templates are used to
+    # generate the repo urls used during maintenance update testing
+    MAINTENANCE_REPO_TEMPLATES = {
+        'ses6': {
+            '{}-update':
+                'http://download.suse.de/ibs/SUSE:/Maintenance:/{}/SUSE_SLE-15_Update',
+            '{}-update-sp1':
+                'http://download.suse.de/ibs/SUSE:/Maintenance:/{}/SUSE_SLE-15-SP1_Update',
+            '{}-storage':
+                'http://download.suse.de/ibs/SUSE:/Maintenance:/{}/SUSE_Updates_Storage_6_x86_64'
+        },
+        'ses7': {
+            '{}-update':
+                'http://download.suse.de/ibs/SUSE:/Maintenance:/{}/SUSE_SLE-15_Update',
+            '{}-update-sp2':
+                'http://download.suse.de/ibs/SUSE:/Maintenance:/{}/SUSE_SLE-15-SP2_Update',
+            '{}-storage':
+                'http://download.suse.de/ibs/SUSE:/Maintenance:/{}/SUSE_Updates_Storage_7_x86_64'
+        },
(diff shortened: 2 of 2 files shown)


sesdev/__init__.py
```diff
@@ -1642,3 +1642,53 @@ def shell_completion(shell):
...
+@click.argument('version')
+@common_create_options
+@deepsea_options
+@ceph_salt_options
+@libvirt_options
+@ipv6_options
+def maintenance_test(maintenance_id, version, **kwargs):
+    """
+    Create a test cluster for installation/smoke testing of package maintenance
+    updates. Takes a maintenance id of the form 'SUSE:Maintenance:xxxxx:yyyyyy'
```
kshtsk, line 1658: > SUSE:Maintenance:xxxxx:yyyyyy
what this value is reality? like IBS repo? Or some well known term `maintenance id`, if so I would like to see a reference where it is defined.

sesdev/__init__.py
```diff
@@ -1642,3 +1642,53 @@ def shell_completion(shell):
...
+    or 'S:M:xxxxx:yyyyyy' and a SES version and creates a test cluster.
+
+    The difference between the test cluster and a 'normal' sesdev cluster is
```
kshtsk, line 1661: Which corresponding commands do mean, maybe it is worth to mention them here as a reference?

sesdev/__init__.py
```diff
@@ -1642,3 +1642,53 @@ def shell_completion(shell):
...
+    that the test cluster has exra repositories from which it installs the
+    packages under testing.
+    """
+    (_, _, _mi, _rr) = maintenance_id.split(':')
```
kshtsk, line 1665: (style)
1. It is not a good idea to local variables with _ (underscore).
2. what if the maintenance_id str does not match the template? maybe we want to handle this and report user what do we expect as an argument here and quit?

sesdev/__init__.py
```diff
@@ -1642,3 +1642,53 @@ def shell_completion(shell):
...
+    deployment_id = "{}-{}-{}".format(_mi, _rr, version)
+
+    vnums = {
+        'ses6': '6',
+        'ses7': '7',
+        'ses7p': '7.1'
+    }
+    vnum = vnums[version]
```
kshtsk, line 1673: What if we got the key error here? I think we need to report user that we do not support the given version yet and quit?

kshtsk (commented):
see the comments above

sesdev/__init__.py
```diff
@@ -1642,3 +1642,75 @@ def shell_completion(shell):
...
+    def _parse_review_request_id(rrid):
+        try:
+            _, _, xmi, xrr = rrid.split(':')
+            return (xmi, xrr)
+        except ValueError:
+            click.echo(f"Review Request ID {rrid} does not conform to pattern S:M:xxxxx:yyyyyy")
+            sys.exit(1)
+
+    xmi, xrr = _parse_review_request_id(review_request_id)
+    deployment_id = f"{xmi}-{xrr}-{version}"
```
kshtsk, line 1694: Did you look at another similar commands? There is a usage of _create_command but with prior `_maybe_gen_dep_id` function, so maybe use similar approach here, having a dedicated function for creating id. Also, have you tested the _destroy_ command with the cluster created for maintenance testing? It is using `_maybe_glob_deps` for matching deployments, did it work correctly?

sesdev/__init__.py
```diff
@@ -1642,3 +1642,75 @@ def shell_completion(shell):
...
+
+@cli.command()
+@click.argument('review_request_id')
+@click.argument('version')
+@common_create_options
+@deepsea_options
+@ceph_salt_options
+@libvirt_options
+@ipv6_options
+def maintenance_test(review_request_id, version, **kwargs):
```
kshtsk, line 1655: I would prefer shorter naming for the command. But let's discuss a bit of general style of the sesdev. As far as I understand this command should be similar or synonym to the `create` command like the others ses5, ses6, etc.
Maybe, maybe... there is more sense to add a common option to other commands, like `--maintenance` where one can provide an incident id?

sesdev/__init__.py
```diff
@@ -1642,3 +1642,75 @@ def shell_completion(shell):
...
+    """
+    Create a test cluster for installation/smoke testing of package maintenance
+    updates. Takes a review request id of the form 'SUSE:Maintenance:xxxxx:yyyyyy'
+    or 'S:M:xxxxx:yyyyyy' and a SES version and creates a test cluster.
```
kshtsk, line 1659: I suggest to give a reader idea where this term comes from by providing a link to the documentation, like, https://gitlab.suse.de/qa-maintenance/mtui/-/blob/master/Documentation/cli.rst or you have better something in mind?

kshtsk: And use RRID in the same way.

sesdev/__init__.py
m-ildefons: Yes, `destroy` works correctly. The reason for not using `_maybe_gen_dep_id` is that there is a need to make sure that a cluster can be identified 1:1 to its corresponding review request, otherwise mix ups could happen, polluting the test results.
Since a cluster might be kept around for as long as the maintenance incident lasts and a single maintenance incident can have multiple review requests and a review may be necessary on multiple versions, it is a good idea make it possible to tie a cluster directly to the review and version.

sesdev/__init__.py
m-ildefons: The whole point of this command is to provide a **short hand** for this specific kind of cluster creation. Consolidating this functionality and creating a more general command with more options would do the exact opposite.
It makes a lot of sense to have it as separate command. Please don't force me to write custom shell wrappers for sesdev, because sesdev is just a wrapper for Vagrant.

kshtsk: > The whole point of this command is to provide a **short hand** for this specific kind of cluster creation. Consolidating this functionality and creating a more general command with more options would do the exact opposite. It makes a lot of sense to have it as separate command. 

At this point, I don't see that many options, just two or one extra, and we don't want to cover all combinations right now. Take a look at: instead of writing `sesdev maintenance_test RRID --version ses5` you just use `sesdev ses5 DEPL --maintenance RRID`.


> Please don't force me to write custom shell wrappers for sesdev, because sesdev is just a wrapper for Vagrant.

Where did I say you must make shell wrappers? :-) please don't

tserong (commented):
The one thing I'd suggest adding to this is to somehow make it use the devel container images as listed in IMAGE_PATHS_DEVEL in seslib/constant.py, at least until we figure out how to get a proper ceph container image out of an MU under test. Otherwise, this'll be testing the released container images from registry.suse.com.

sesdev/__init__.py
```diff
@@ -1642,3 +1643,105 @@ def shell_completion(shell):
...
+            return (xmi, xrr)
+        except ValueError:
+            click.echo(f"Review Request ID {rrid} does not conform to pattern S:M:xxxxx:yyyyyy")
+            sys.exit(1)
+
+    xmi, xrr = _parse_review_request_id(review_request_id)
+    deployment_id = f"{xmi}-{xrr}-{version}"
+
+    vnum = _version_to_number(version)
+    sp = _version_to_sp(version)
```
kshtsk, line 1724: (suggestion) Maybe reorganize the code to have a single mapping of SES version to SLE version?

sesdev/__init__.py
```diff
@@ -1642,3 +1643,105 @@ def shell_completion(shell):
...
+
+            if get.status_code == 200:
+                return True
+            return False
+        except requests.exceptions.RequestException:
+            return False
+
+    def _add_url(url_dict, name, url):
+        if _check_url(url):
+            url_dict[name] = url
```
kshtsk, line 1686: (suggestion) this is a bit a complication to have this method, instead construct a dict of repos, and then use comprehension.

Let say you've got a list of repo urls:
```
repo_candidates = dict(repo1=url1, repo2=url2, ...)
available_repos = [k:v for k,v in repo_candidates.items() if _check_url(v)]
```

sesdev/__init__.py
```diff
@@ -1642,3 +1643,105 @@ def shell_completion(shell):
...
+    vnum = _version_to_number(version)
+    sp = _version_to_sp(version)
+
+    repo_url = "http://download.suse.de/ibs/SUSE:/Maintenance:"
+
+    repo_urls = {}
+
+    _add_url(repo_urls, f"{xmi}-updates", f"{repo_url}/{xmi}/SUSE_SLE-15_Update")
+    _add_url(repo_urls, f"{xmi}-updates-{sp}", f"{repo_url}/{xmi}/SUSE_SLE-15-{sp}_Update")
+    _add_url(repo_urls, f"{xmi}-storage", f"{repo_url}/{xmi}/SUSE_Updates_Storage_{vnum}_x86_64")
```
kshtsk, line 1732: For better readability, maybe use dictionary of repos, with probable a placeholder just for single maintenance incident id?

sesdev/__init__.py
m-ildefons: This is mapping
a) our internal names for SES to the official version numbers **of SES**
b) our internal names for SES to the respective SLES versions
No amount of reorganization is going to unify that.

sesdev/__init__.py
m-ildefons: I strongly oppose the use of a placeholder with the inevitable special casing everywhere.

m-ildefons: I'm not sure what kind of change you want to see for this code.

m-ildefons:
> The one thing I'd suggest adding to this is to somehow make it use the devel container images as listed in IMAGE_PATHS_DEVEL in seslib/constant.py, at least until we figure out how to get a proper ceph container image out of an MU under test. Otherwise, this'll be testing the released container images from registry.suse.com.

The reason I don't want to do this is that in 90% of test cases packages other than ceph, ceph-salt etc. are under test and the test just makes sure that ceph & co. doesn't break from the update to the other packages. It's therefore a good thing to test against the same containers we deliver to customers. Only when we release a MU to ceph itself this becomes a problem and in that case the command takes `--ceph-image-path` and similar arguments just like the `create` subcommand to change the container under test.

The proper solution to this is of course to build and deliver the MUs as containers as well at a predictable URL, so they can be tested, but until then one would have to know to test against the devel containers anyways.

sesdev/__init__.py
kshtsk: I want this, for example:
```
repos = {
   'ses6': {
      'sle15up': f'http://download.suse.de/ibs/SUSE:/Maintenance:{incident_id}/SUSE_SLE-15_Update',
      'sle15sp1_update': f'http://download.suse.de/ibs/SUSE:/Maintenance:{incident_id}/SUSE_SLE-15-SP1_Update',
      'ses6': f'http://download.suse.de/ibs/SUSE:/Maintenance:{incident_id}/SUSE_Updates_Storage_6_x86_64',
   },
   'ses7': {
      'sle15up': f'http://download.suse.de/ibs/SUSE:/Maintenance:{incident_id}/SUSE_SLE-15_Update',
      'sle15sp2_update': f'http://download.suse.de/ibs/SUSE:/Maintenance:{incident_id}/SUSE_SLE-15-SP2_Update',
      'ses7': f'http://download.suse.de/ibs/SUSE:/Maintenance:{incident_id}/SUSE_Updates_Storage_7_x86_64',
   },
   .
   .
   .
}
```

It is way easier to maintain the code than when you programatically define each bit of url.

sesdev/__init__.py
kshtsk: I want this, for example:
```
repos = {
   'ses6': {
      'sle15up': f'http://download.suse.de/ibs/SUSE:/Maintenance:{incident_id}/SUSE_SLE-15_Update',
      'sle15sp1_update': f'http://download.suse.de/ibs/SUSE:/Maintenance:{incident_id}/SUSE_SLE-15-SP1_Update',
      'ses6': f'http://download.suse.de/ibs/SUSE:/Maintenance:{incident_id}/SUSE_Updates_Storage_6_x86_64',
   },
   'ses7': {
      'sle15up': f'http://download.suse.de/ibs/SUSE:/Maintenance:{incident_id}/SUSE_SLE-15_Update',
      'sle15sp2_update': f'http://download.suse.de/ibs/SUSE:/Maintenance:{incident_id}/SUSE_SLE-15-SP2_Update',
      'ses7': f'http://download.suse.de/ibs/SUSE:/Maintenance:{incident_id}/SUSE_Updates_Storage_7_x86_64',
   },
   .
   .
   .
}
```

It is way easier to maintain the code than when you programatically define each bit of url.

sesdev/__init__.py
kshtsk: @tserong If you're okay with this, please give a review mark to the PR, so I do not take the responsibility to the solution I am not fine with ;-)

tserong:
> The reason I don't want to do this is that in 90% of test cases packages other than ceph, ceph-salt etc. are under test and the test just makes sure that ceph & co. doesn't break from the update to the other packages.

Ah, of course. Makes perfect sense.

sesdev/__init__.py
m-ildefons: Ok I misunderstood you there first. This makes a lot of sense.

sesdev/__init__.py
```diff
@@ -1642,3 +1643,81 @@ def shell_completion(shell):
...
+        try:
+            get = requests.get(url)
+
+            if get.status_code == 200:
+                return True
+            return False
+        except requests.exceptions.RequestException:
+            return False
+
+    def _parse_review_request_id(rrid):
```
kshtsk, line 1684: Function lack description like: _Returns a pair of incident id and request id parsed from the Review Request ID_

sesdev/__init__.py
```diff
@@ -1642,3 +1643,81 @@ def shell_completion(shell):
...
+        try:
+            _, _, xmi, xrr = rrid.split(':')
```
kshtsk, line 1686: I guess the function does not care about what the first words and it will not signal about an error if you give `::1234:-1`

tserong (approved):
Probably wants rebasing now, but otherwise LGTM, assuming that last commit has taken care of @kshtsk's remaining suggestions.

PR #660, opened 2022-07-21
ses5: remove ses5 capabilities

m-ildefons:
Remove SES 5 capabilities from sesdev and scripts. SES 5 is so far
unsupported that the special casing and the maintenance burdon of the
additional code does no favors.

With SES 5 capabilities, a lot of specific code can be removed from
templates, settings and quite a few special cases around things like
node roles can be removed too, since they become meaningless.
This applies in particular to code around OpenAttic too.


kshtsk:
I am not sure about this, but don't we have customers with ses5 still? @asettle

asettle:
> I am not sure about this, but don't we have customers with ses5 still? @asettle

We do, but we need to wind down the support here. We haven't had any issues for months and this is costing us time and effort. Why would we keep this running at this stage?

kshtsk:
> > I am not sure about this, but don't we have customers with ses5 still? @asettle
> 
> We do, but we need to wind down the support here. We haven't had any issues for months and this is costing us time and effort. Why would we keep this running at this stage?

If you promise not to ask me deploy an ses5 instance for reproduce or something in the future, I would merge this PR with satisfaction to the greatest extent 😿

tserong:
Can I suggest tagging whichever commit is just before this one when it gets merged - call it "final-ses5" or whatever (I don't really care) - so that if we ever do need to do another ses5 deployment we can easily checkout the last sesdev release that was capable of deploying ses5.

Realistically the only time I see us needing to deploy ses5 in future is if we have to reproduce some bizarre issue that only happens on systems that were upgraded from 5 to 6 to 7 (or 7.1). We're never going to need ses5 deployments again for any sort of normal/regular testing.

asettle:
> Can I suggest tagging whichever commit is just before this one when it gets merged - call it "final-ses5" or whatever (I don't really care) - so that if we ever do need to do another ses5 deployment we can easily checkout the last sesdev version that was capable of deploying ses5.
> 
> Realistically the only time I see us needing to deploy ses5 in future is if we have to reproduce some bizarre issue that only happens on systems that were upgraded from 5 to 6 to 7 (or 7.1). We're never going to need ses5 deployments again for any sort of normal/regular testing.

Great suggestion, @m-ildefons could you do that, please? Tim is right. The only time we'd *ever* touch something for SES 5 is related to upgrades. Even so, most customers on SES 5 will be unlikely to upgrade, but we don't want to shoot ourselves in the foot.

tserong:
> I'd love to add the tag, but I don't have permissions to push a tag disappointed

Weird. I assumed any of us in the storage developers team should be able to do that. Maybe let's just get this merged first, then we can figure out who's got tagging permissions and tag this commit minus one after the merge?

kshtsk (approved):

tserong:
Apparently I can do tags - I've just pushed a tag for 941a703f0304a13b00037025a41b3f1b6def4880 as "ses5-final", which is the merge commit before this PR's merge commit.

m-ildefons:
Fantastic :+1: Thanks a lot!

PR #662, opened 2022-08-02
jenkins: fix ci and add leap 15.4

kshtsk:
Here we drop https://download.opensuse.org/repositories/Virtualization:/vagrant repo because from leap 15.3 vagrant-libvirt is included in stock version.
Also we drop removing conflicting packages because it is not necessary anymore. 

Signed-off-by: Kyr Shatskyy <kyrylo.shatskyy@suse.com>


m-ildefons (approved):
LGTM

PR #663, opened 2022-08-08
install: install updates from custom repos

m-ildefons:
Install updates from custom repos, if any are available, for all
packages that are installed before starting the installation procedure
for ceph.
This ensures that when custom repos provide e.g. new versions of salt
or zypper those versions are installed before the cluster is installed.
This functionality is useful for maintenance tests against packages used
during the installation of ceph.


kshtsk (approved):
lgtm

PR #665, opened 2022-08-09
maintenance report: include ceph versions

m-ildefons:
Include ceph version information in the maintenance report. This is
useful to ensure that at the end of the test workflow the versions
running in the cluster (both package- and container-versions) are as
expected.

### DIFF
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -1361,6 +1361,21 @@ def report(deployment_id):
         for pkg in pkgs:
             click.echo(_indent(f"{pkg.name}  {pkg.version}  (from {pkg.repo})"))
     click.echo("")
+
+    click.echo("* Ceph versions:")
+    click.echo("")
+    versions = dep.list_versions()
+    click.echo(_indent("service type | count | version"))
+    for service_name in versions.keys():
+        service_name_is_already_printed = False
+        for service_version, service_count in versions[service_name].items():
+            if service_name_is_already_printed:
+                click.echo(_indent(f"             | {service_count:5} | {service_version}"))
+            else:
+                click.echo(_indent(f"{service_name:12} | {service_count:5} | {service_version}"))
+                service_name_is_already_printed = True
+    click.echo("")
+
     click.echo(f"* {ver} cluster status:")
     click.echo("")
     dep.ssh('master', ['ceph', 'status'], interactive=False)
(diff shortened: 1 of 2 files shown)


sesdev/__init__.py
```diff
@@ -1361,6 +1361,21 @@ def _indent(string, spaces=4):
         for pkg in pkgs:
             click.echo(_indent(f"{pkg.name}  {pkg.version}  (from {pkg.repo})"))
     click.echo("")
+
+    click.echo("* Ceph versions:")
+    click.echo("")
+    versions = dep.list_versions()
+    click.echo(_indent("service type | count | versions"))
+    for svc in versions.keys():
+        svcp = False
```
kshtsk, line 1370: (style) Can we have less cryptic variable names? The reason, variable svc, svcp, svcver, svcnt has very similar names and this make the code very hard to read. When you want short naming in a small context (like cycles or loops) it is better to give single letter names, it makes more distinguishable, or give a single word name.

sesdev/__init__.py
```diff
@@ -1361,6 +1361,21 @@ def _indent(string, spaces=4):
...
+        for svcver, svccnt in versions[svc].items():
+            if not svcp:
```
kshtsk, line 1372: It is better avoid negative conditions in the code.

sesdev/__init__.py
```diff
@@ -1361,6 +1361,21 @@ def _indent(string, spaces=4):
         for pkg in pkgs:
             click.echo(_indent(f"{pkg.name}  {pkg.version}  (from {pkg.repo})"))
     click.echo("")
+
+    click.echo("* Ceph versions:")
+    click.echo("")
+    versions = dep.list_versions()
+    click.echo(_indent("service type | count | versions"))
```
kshtsk, line 1368: I don't know how the printed table should look like, but is the last column really should be 'versions' or it is just a 'version', i.e. is it multiple version or single per line?

seslib/deployment.py
```diff
@@ -1866,6 +1866,19 @@ def _to_package_list(raw):
 
         return packages
 
+    def list_versions(self):
+        """
+        Returns Ceph version information as returned by `ceph versions` in a
+        python dict
+        """
+        cmd = "ceph versions"
+        node = list(self.nodes.keys())[0]
```
kshtsk, line 1875: just wonder do we care about the first node only, if so maybe give a notice in the function description?

m-ildefons: We only need to execute that command on one node. Any node would do, but the first one is guaranteed to exist.

kshtsk (approved):

PR #666, opened 2022-08-10
maintenance report: add list of repos

m-ildefons:
Add list of configured repos to the maintenance report. This is useful
to trace back which repo url a specific package listed in the same
maintenance report came from. It will also be useful in the future to
ensure a package update test actually installed the expected packages
and not packages from a different repo.

### DIFF
--- a/seslib/deployment.py
+++ b/seslib/deployment.py
@@ -1806,6 +1806,36 @@ deployment might not be completely destroyed.
             ssh_cmd.append("ceph mgr module enable {}".format(module))
             tools.run_sync(ssh_cmd)
 
+    def list_repos(self):
+        """
+        Compile a list of zypper repos configured on each node of the cluster.
+        Returns a dictionary by node name to the list of repos configured on
+        that node.
+        """
+        def _to_repo_list(raw):
+            lines = raw.splitlines()
+            repo_list = []
+
+            for line in lines:
+                cols = line.split('|')
+                if len(cols) > 1 and cols[0] != '# ':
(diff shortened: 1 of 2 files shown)


seslib/deployment.py
```diff
@@ -1806,6 +1806,36 @@ def replace_mgr_modules(self, local=None, pr=None, branch=None, repo=None, langs
             ssh_cmd.append("ceph mgr module enable {}".format(module))
             tools.run_sync(ssh_cmd)
 
+    def list_repos(self):
+        """
+        Compile a list of zypper repos configured on each node of the cluster.
+        Returns a dictionary by node name to the list of repos configured on
```
kshtsk, line 1812: "by"?

seslib/deployment.py
```diff
@@ -1806,6 +1806,36 @@ def replace_mgr_modules(self, local=None, pr=None, branch=None, repo=None, langs
             ssh_cmd.append("ceph mgr module enable {}".format(module))
             tools.run_sync(ssh_cmd)
 
+    def list_repos(self):
+        """
+        Compile a list of zypper repos configured on each node of the cluster.
```
kshtsk, line 1811: I guess this line is extra.

kshtsk (approved):

seslib/deployment.py
m-ildefons: "by node name" means in this context that the node names are the keys in the dictionary and the repos are the values. This isn't obvious since the mapping could also be done by repo alias or some other way, but doing it this way accounts for the possibility of having different repos on different nodes and makes it much easier to iterate through the entire cluster.

Issue #672, opened 2022-12-15
I can no longer deploy a SES7p environment


s0nea:
If I do `sesdev create ses7p --single-node ses7p-mini`, it's resulting in:

```
master: Failure in minion: master.ses7p-mini.test
    master: __id__: install cephadm
    master: __run_num__: 72
    master: __sls__: ceph-salt.apply.cephtools
    master: changes: {}
    master: comment: "An error was encountered while installing package(s): Zypper command failure:\
    master:   \ Running scope as unit: run-re1b27466ef274ec3956bcb42a502d618.scopeLoading repository\
    master:   \ data...\nReading installed packages...\nResolving package dependencies...\n\n\
    master:   Problem: the to be installed ceph-base-16.2.9.536+g41a9f9a5573-150300.3.3.1.x86_64\
    master:   \ requires 'ceph-common = 16.2.9.536+g41a9f9a5573-150300.3.3.1', but this requirement\
    master:   \ cannot be provided\n  not installable providers: ceph-common-16.2.9.536+g41a9f9a5573-150300.3.3.1.x86_64[storage-update]\n\
    master:   \ Solution 1: do not install ceph-common-16.2.9.536+g41a9f9a5573-150300.6.3.1.x86_64\n\
    master:   \ Solution 2: do not install ceph-base-16.2.9.536+g41a9f9a5573-150300.3.3.1.x86_64\n\
    master:   \ Solution 3: break ceph-base-16.2.9.536+g41a9f9a5573-150300.3.3.1.x86_64 by ignoring\
    master:   \ some of its dependencies\n\nChoose from above solutions by number or cancel [1/2/3/c/d/?]\
    master:   \ (c): c"
    master: duration: 725.412
    master: name: install cephadm
    master: result: false
    master: start_time: '10:05:03.261770'
    master: state: pkg_|-install cephadm_|-install cephadm_|-installed
```

tserong:
This is happening because of an unfortunate set of circumstances:

1) There's currently a slight version discrepancy between the ceph-common package present in the SLE 15 SP3 Basesystem vs. SES 7.1 Updates repositories. SP3 Updates has ceph-common-16.2.9.536+g41a9f9a5573-150300.6.3.1, whiles SES 7.1 Updates has ceph-common-16.2.9.536+g41a9f9a5573-150300.3.3.1 (note: 6.3.1 vs. 3.3.1 at the end - it's exactly the same code, just a different build or release number generated by the SUSE build service).
[...]

PR #677, opened 2022-12-21
qa: fix make_salt_master_an_admin_node_test and NFS tests

tserong:
This is related to https://github.com/SUSE/sesdev/issues/672. When running `sesdev create ses7p --qa-test`, the
make_salt_master_an_admin_node_test function will try to install ceph-common.  If we're suffering from subtle version drift between ceph-common and ceph-base from different repos, this will fail with something like:
```
  Problem: the installed ceph-base-16.2.9.536+g41a9f9a5573-150300.3.3.1.x86_64 requires 'ceph-common = 16.2.9.536+g41a9f9a5573-150300.3.3.1', but this requirement cannot be provided
   Solution 1: deinstallation of ceph-base-16.2.9.536+g41a9f9a5573-150300.3.3.1.x86_64
   Solution 2: do not install ceph-common-16.2.9.536+g41a9f9a5573-150300.6.3.1.x86_64
   Solution 3: break ceph-base-16.2.9.536+g41a9f9a5573-150300.3.3.1.x86_64 by ignoring some of its dependencies
```
This is because zypper is trying to upgrade us to a newer version of ceph-common from a different repo.  We can avoid this problem by not trying to install ceph-common if it's already present.

### DIFF
--- a/qa/common/common.sh
+++ b/qa/common/common.sh
@@ -1094,8 +1094,12 @@ function nfs_maybe_list_objects_in_recovery_pool_test {
         if [ "$NFS_NODE_LIST" ] && [ "$MDS_NODE_LIST" ] ; then
             # NFS Recovery Pool expected to exist
             skipped=""
+            local nfs_pool=".nfs"
+            if [ "$DEPLOYMENT_VERSION" = "octopus" ] ; then
+                nfs_pool="nfs-ganesha"
+            fi
             set -x
-            rados --pool nfs-ganesha --namespace sesdev_nfs ls | tee "$tmpfile"
+            rados --pool "$nfs_pool" --namespace sesdev_nfs ls | tee "$tmpfile"
             set +x
             if [ -s "$tmpfile" ] ; then
                 result="OK"
@@ -1126,7 +1130,11 @@ function nfs_maybe_create_export {
         if [ "$NFS_NODE_LIST" ] && [ "$MDS_NODE_LIST" ] ; then
             skipped=""
             set -x
-            ceph nfs export create cephfs sesdev_fs sesdev_nfs "/sesdev_nfs"
+            if [ "$DEPLOYMENT_VERSION" = "octopus" ] ; then
+                ceph nfs export create cephfs sesdev_fs sesdev_nfs "/sesdev_nfs"
+            else
+                ceph nfs export create cephfs sesdev_nfs "/sesdev_nfs" sesdev_fs
+            fi
             ceph nfs export ls sesdev_nfs --detailed | tee "$tmpfile"
             length="$(<"$tmpfile" jq -r 'length')"
             set +x
--- a/qa/common/common.sh
+++ b/qa/common/common.sh
@@ -218,7 +218,7 @@ function make_salt_master_an_admin_node_test {
     local arbitrary_mon_node
     echo
     echo "WWWW: make_salt_master_an_admin_node_test"
-    _zypper_install_on_master ceph-common
+    rpm -q --quiet ceph-common || _zypper_install_on_master ceph-common
     set -x
     mkdir -p "/etc/ceph"
     if [ -f "$ADMIN_KEYRING" ] || [ -f "$CEPH_CONF" ] ; then


tserong:
I found some more broken stuff when running NFS tests. I figured adding that to this PR was more straightforward than opening another PR, but LMK if you'd rather I split these up.

kshtsk:
@tserong maybe rebase it?

kshtsk:
@tserong one more time, please

kshtsk:
Looks ceph filesystem repos for octopus are gone on obs

tserong:
> Looks ceph filesystem repos for octopus are gone on obs

Looks like that's due to this request: https://build.opensuse.org/request/show/1049229

We could re-enable Leap 15.2 builds for filesystems:ceph:octopus, but I'm struggling to see why we would want to.

kshtsk:
> > Looks ceph filesystem repos for octopus are gone on obs
> 
> Looks like that's due to this request: https://build.opensuse.org/request/show/1049229
> 
> We could re-enable Leap 15.2 builds for filesystems:ceph:octopus, but I'm struggling to see why we would want to.

Right, it's because auto accept applied, now the jenkins job should be changed to the higher version of ceph? Which of it does have tumbleweed and leap repos?

tserong:
> Right, it's because auto accept applied, now the jenkins job should be changed to the higher version of ceph? Which of it does have tumbleweed and leap repos?

Ah, OK. I'd missed that jenkins job :-/  I'm happy to re-enable octopus builds against Leap 15.2 on OBS if you'd like, I'm just not sure if it's worth continuing to test octopus deployments, given octopus reached end of life on 2022-08-09 (see https://docs.ceph.com/en/latest/releases/#archived-releases)

We've still got Ceph Pacific building on OBS in filesystems:ceph:pacific, for Leap 15.3, 15.4 and Tumbleweed, but container images stopped building there because of another one of those auto-accepted cleanup requests: https://build.opensuse.org/request/show/1051452. I've since added the images repo back in, so once everything's build again, _Pacific_ should be fine.

m-ildefons:
Seems like the repo for Ceph Octopus isn't available anymore:
https://download.opensuse.org/repositories/filesystems:/ceph:/octopus/openSUSE_Leap_15.2
Which is why the Jenkins job is failing. Is this intentional?
In any case, I think we may do well to move that to Pacific anyways, as that is the version we care about going forward.

s0nea:
> Seems like the repo for Ceph Octopus isn't available anymore: https://download.opensuse.org/repositories/filesystems:/ceph:/octopus/openSUSE_Leap_15.2 Which is why the Jenkins job is failing. Is this intentional?

It seems like all build targets have been removed from https://build.opensuse.org/project/show/filesystems:ceph:octopus

m-ildefons:
That'd probably be because Leap 15.2 is EOLed along with SLE 15 SP2.

kshtsk:
@m-ildefons see the comments above

kshtsk:
@tserong any progress with repo issues?

kshtsk:
@tserong Btw, if we switch to defaults pacific, then it means we have to drop octopus support, doesn't it?

tserong:
Rebased.

> @tserong Btw, if we switch to defaults pacific, then it means we have to drop octopus support, doesn't it?

Having switched the default to pacific, it just means we're not testing octopus anymore. But that's OK, octopus is EOL now anyway.

trociny:
Ah, I did not noticed this PR and had my patch for this too. I did it a little differently though, because I  am not sure using `DEPLOYMENT_VERSION` is the best approach here. I think it will return `octopus` only for `ceph create octopus`, but if you run `ceph create ses7` it will return `ses7`, although it is "octopus" too. So I used ceph version instead:
```
if [[ "$(ceph --version)" < "ceph version 16" ]]; then
    ...
```
And the nfs pool could be found without checking the version, something like:
```
            pool="$(rados lspools | grep nfs)"
            test "$pool" = nfs-ganesha -o "$pool" = .nfs
```

PR #680, opened 2023-01-12
worklfows/tox: update action uses

kshtsk:
This fixes python tox and lint failures, also we drop 3.8 since it cannot be found anymore in Leap 15.4


m-ildefons (approved):
LGTM

PR #684, opened 2023-03-13
qa: add iscsi test

trociny:


kshtsk:
the test is failing on jenkins

trociny:
> the test is failing on jenkins

I believe I have fixed the issue. Will it rerun on jenkins automatically after I have pushed the update?

kshtsk:
> > the test is failing on jenkins
> 
> I believe I have fixed the issue. Will it rerun on jenkins automatically after I have pushed the update?

Just do it and see

trociny:
jenkins failed but this time it does not look related

tserong (approved):

tserong:
> jenkins failed but this time it does not look related

Yeah, looks like a missing admin keyring after deployment, which is unrelated, makes no sense, and I can't reproduce.

tserong:
...and now having looked in more detail: #690 should fix that.

tserong:
> But this is an example, why we shouldn't merge a PR without all tests passed.

Indeed.

PR #685, opened 2023-03-15
k3s: make sesdev tunnel k3s dashboard work 

trociny:


trociny:
Additionally, it would be nice if the dashboard admin password matched the non-k3s deployment dashboard password ("admin"), but right now I have no idea how to make it.

It could be done by running something like this after deployment:
```
kubectl -n rook-ceph exec -it deploy/rook-ceph-tools --  ceph dashboard  ac-user-set-password admin -i <file_with_passwd>  --force-password
```
But for this to work we should wait for the tool container to deploy, which we currently don't.

tserong:
I tried running this, but (at least in my testing) the rook-ceph-mgr-dashboard service wasn't up yet, so it just fell through and didn't enable the tunnel. I guess we really need to wait for at least partial deployment (mons, mgrs) somehow?

tserong:
I experimented with adding `kubectl wait --timeout=5m --namespace rook-ceph '--for=jsonpath={.spec.type}="ClusterIP"' service/rook-ceph-mgr-dashboard`, which I know is a bit sketchy -- this is just waiting for the spec to appear, because services don't support `--condition=ready` (see https://github.com/kubernetes/kubernetes/issues/80828#issuecomment-549821445) -- but it still didn't work, because initially the service does't even exist, so `kubectl wait` fails immediately with `Error from server (NotFound): services "rook-ceph-mgr-dashboard" not found`.

trociny:
> I tried running this, but (at least in my testing) the rook-ceph-mgr-dashboard service wasn't up yet, so it just fell through and didn't enable the tunnel. I guess we really need to wait for at least partial deployment (mons, mgrs) somehow?

Did it work when the rook-ceph-mgr-dashboard service became up (i.e. for the case this patch is supposed to work)? I think it is ok to have the tunnel command to fail when the service is not up yet. We may want though to improve error messaging/handling if the current current one does not look nice.

tserong:
> Did it work when the rook-ceph-mgr-dashboard service became up (i.e. for the case this patch is supposed to work)?

It took a while :-) but yes it did work fine once the service came up.

> I think it is ok to have the tunnel command to fail when the service is not up yet. We may want though to improve error messaging/handling if the current current one does not look nice.

Yeah, makes sense. Let's merge this and worry about messaging separately if necessary.

PR #686, opened 2023-03-16
k3s: default OS to sles-15-sp3 if --deploy-ses is specified

tserong:


tserong:
ping @trociny

tserong:
Closing after discussion in #692 (TL;DR: it's weird/unexpected to have `--deploy-*` options change the default OS)

PR #692, opened 2023-04-19
Add ability to deploy Longhorn on k3s

tserong:
This adds the `--deploy-longhorn` option to `sesdev create k3s`, which will make sesdev deploy Longhorn instead of Ceph.  By default it will deploy the latest released version of Longhorn on SLE 15 SP3, but this can be overridden by using the `--longhorn-version` and `--os options` respectively.
    
When deploying Longhorn, one extra 8GB disk will be added to each worker node mounted as ext4 under /var/lib/longhorn.  One possible area for future work is to make it use multiple disks per node.
    
The Longhorn UI will be made available via HTTP with no authentication.

### DIFF
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -964,7 +964,7 @@ def caasp4(deployment_id, **kwargs):
 @click.option("--deploy-longhorn", is_flag=True, default=False,
               help="Deploy Longhorn on k3s")
 @click.option("--longhorn-version", default=None,
-              help='Longhorn version to install, e.g. "v1.4.1" (defaults to latest stable)')
+              help='Longhorn version to install, e.g. "1.4.1" (defaults to latest stable)')
 @click.option("--k3s-version", default=None,
               help='k3s version to install (defaults to latest stable)')
 def k3s(deployment_id, **kwargs):
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -387,6 +387,7 @@ def _gen_settings_dict(
         deepsea_branch=None,
         deepsea_repo=None,
         deploy_ses=None,
+        deploy_longhorn=None,
         devel=None,
         disk_size=None,
         dry_run=None,
@@ -437,6 +438,7 @@ def _gen_settings_dict(
         msgr2_secure_mode=None,
         msgr2_prefer_secure=None,
         k3s_version=None,
+        longhorn_version=None,
 ):
 
     settings_dict = {}
@@ -665,6 +667,9 @@ def _gen_settings_dict(
     if deploy_ses and version == 'k3s':
         settings_dict['k3s_deploy_ses'] = True
 
+    if deploy_longhorn and version == 'k3s':
+        settings_dict['k3s_deploy_longhorn'] = True
+
     for folder in synced_folder:
         try:
             src, dst = folder.split(':')
@@ -704,6 +709,9 @@ def _gen_settings_dict(
 
     if k3s_version is not None:
         settings_dict['k3s_version'] = k3s_version
+
+    if longhorn_version is not None:
+        settings_dict['longhorn_version'] = longhorn_version
 
     return settings_dict
 
@@ -790,6 +798,17 @@ def _create_command(deployment_id, settings_dict):
                     click.echo("  # kubectl -n rook-ceph exec -it deploy/rook-ceph-tools -- bash")
                     click.echo()
                     click.echo("Inside the toolbox you can use the ceph CLI (`ceph status` etc.)")
+                    click.echo()
+                if dep.settings.k3s_deploy_longhorn:
+                    click.echo("Longhorn will now be deploying, which may take some time.")
+                    click.echo("After logging into the cluster, try these:")
+                    click.echo()
+                    click.echo("  # kubectl get pods -n longhorn-system --watch")
+                    click.echo("  # kubectl get pods -n longhorn-system")
+                    click.echo()
+                    click.echo("The Longhorn UI will be accessible via any cluster IP address")
+                    click.echo("(see the `kubectl -n longhorn-system get ingress` output above).")
+                    click.echo("Note that no authentication is required.")
                     click.echo()
             else:
                 click.echo("Or, access the Ceph Dashboard with:")
@@ -942,6 +961,10 @@ def caasp4(deployment_id, **kwargs):
 @libvirt_options
 @click.option("--deploy-ses", is_flag=True, default=False,
               help="Deploy SES using rook in k3s")
+@click.option("--deploy-longhorn", is_flag=True, default=False,
+              help="Deploy Longhorn on k3s")
+@click.option("--longhorn-version", default=None,
+              help='Longhorn version to install, e.g. "v1.4.1" (defaults to latest stable)')
 @click.option("--k3s-version", default=None,
               help='k3s version to install (defaults to latest stable)')
 def k3s(deployment_id, **kwargs):
@@ -953,6 +976,8 @@ def k3s(deployment_id, **kwargs):
     default_dep_id = 'k3s'
     if kwargs['deploy_ses']:
         default_dep_id += '-ses'
+    if kwargs['deploy_longhorn']:
+        default_dep_id += '-longhorn'
     deployment_id = _maybe_gen_dep_id(default_dep_id, deployment_id, settings_dict)
     _create_command(deployment_id, settings_dict)
 
--- a/sesdev/__init__.py
+++ b/sesdev/__init__.py
@@ -950,7 +950,10 @@ def k3s(deployment_id, **kwargs):
     """
     _prep_kwargs(kwargs)
     settings_dict = _gen_settings_dict('k3s', **kwargs)
-    deployment_id = _maybe_gen_dep_id('k3s', deployment_id, settings_dict)
+    default_dep_id = 'k3s'
+    if kwargs['deploy_ses']:
+        default_dep_id += '-ses'
+    deployment_id = _maybe_gen_dep_id(default_dep_id, deployment_id, settings_dict)
     _create_command(deployment_id, settings_dict)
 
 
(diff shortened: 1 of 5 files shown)


sesdev/__init__.py
```diff
@@ -942,15 +961,27 @@ def caasp4(deployment_id, **kwargs):
...
+@click.option("--deploy-longhorn", is_flag=True, default=False,
+              help="Deploy Longhorn on k3s")
+@click.option("--longhorn-version", default=None,
+              help='Longhorn version to install (defaults to latest stable)')
 @click.option("--k3s-version", default=None,
               help='k3s version to install (defaults to latest stable)')
 def k3s(deployment_id, **kwargs):
     """
-    Creates a k3s cluster using Tumbleweed
+    Creates a k3s cluster using Tumbleweed by default, or SLE 15 SP3
```
kshtsk, line 972: the sp3 looks super old, maybe sp5 now?

seslib/templates/k3s/provision.sh.j2
```diff
@@ -115,6 +121,57 @@ kubectl create -f rook-ceph/examples/toolbox.yaml
...
+{% if k3s_deploy_longhorn %}
+
+{% if longhorn_version %}
+longhorn_version={{ longhorn_version }}
+{% else %}
+longhorn_version=$(
+    curl -sL \
+        -H "Accept: application/vnd.github+json" \
+        -H "X-GitHub-Api-Version:2022-11-28" \
+        https://api.github.com/repos/longhorn/longhorn/releases | \
```
kshtsk, line 133: This is really not good idea, since rest api has limited number of requests, as we saw on our CI jobs it may block everything.

m-ildefons (changes requested):
There are two places where I'm not convinced this code is sufficiently robust:
- `sesdev create k3s` without any more options doesn't seem to default to the same base os
- explicitly setting the number of disks may give unexpected results if `--num-disks` is not exactly one.

sesdev/__init__.py
tserong: Well, SP5 isn't released yet...  SP3 is there because I started earlier from the perspective of deploying Ceph via Rook, and in a SES 7.1 context that meant SP3. But you're right, SP3 is getting a bit old.

seslib/templates/k3s/provision.sh.j2
tserong: I'm open to suggestions for another way of finding out the latest longhorn release.

sesdev/__init__.py
```diff
@@ -942,15 +961,27 @@ def caasp4(deployment_id, **kwargs):
...
+    if --deploy-ses or --deploy-longhorn is specified.
     """
     _prep_kwargs(kwargs)
+    if (kwargs['deploy_ses'] or kwargs['deploy_longhorn']) and kwargs['os'] is None:
+        kwargs['os'] = 'sles-15-sp3'
```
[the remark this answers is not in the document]
tserong: Hrm. Maybe I should just not mess with the default os (in which case it will be tumbleweed if not specified) and let the user declare what they want explicitly with `--os`?

seslib/templates/k3s/provision.sh.j2
```diff
@@ -148,4 +205,11 @@ rm /tmp/k3s_token
 
 curl -sfL https://get.k3s.io | K3S_URL=https://{{ master.fqdn }}:6443 sh -
 
+{% if k3s_deploy_longhorn %}
+mkfs.ext4 /dev/vdb
+mkdir /var/lib/longhorn
+echo "/dev/vdb /var/lib/longhorn ext4 defaults 0 2" >> /etc/fstab
```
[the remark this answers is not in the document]
tserong: Here, I hadn't yet figured out how to do something useful with more than one disk. But you're right, my current implementation is a bit sketchy. I'll fiddle with this a bit more.

seslib/templates/k3s/provision.sh.j2
kshtsk: May we have an option for longhorn version as well as k3s version option btw.

kshtsk: And the latest longhorn release is not always the best version to install as far as I understand :-D Maybe this is stupid, but give hint in --help output where to get needed version. I'm not sure if you want to add this running branch to some regular job, we'll see the same as for rookcheck.

tserong: Oh, also: this code path won't even be exercised by the jenkins jobs at the moment anyway, so... Kinda not really a problem yet :-)

kshtsk: well, I don't like the situation when I'd need to debug this MS thing again in our open source project :-D
another stupid suggestion from me, just parse out https://github.com/longhorn/longhorn/tags

kshtsk: Anyway, having this code in remote shell is not good either, just try and do it on python side.

kshtsk: The fact that this code will never run in CI, other cool guys might block the api, you'll have to debug your code trying to understand why curl is failed.

sesdev/__init__.py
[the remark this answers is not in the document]
m-ildefons: Sounds more reasonable, if the default for a 'bare' k3s is to be installed on tumbleweed. Then I'd not expect that adding a `--deploy-*` flag would change the base os to SLES.
I guess my point here is that it would be quite unexpected to have the `--deploy-*` flag implicitly change the base os compared to the base `sesdev create k3s`

seslib/templates/k3s/provision.sh.j2
kshtsk: another reason, why not to do this in shell script, if you are running it on multiple nodes, sequentially it may pick the inconsistent version from git hub, when there is new version appears in the middle

sesdev/__init__.py
```diff
@@ -942,15 +961,27 @@ def caasp4(deployment_id, **kwargs):
 @libvirt_options
 @click.option("--deploy-ses", is_flag=True, default=False,
               help="Deploy SES using rook in k3s")
+@click.option("--deploy-longhorn", is_flag=True, default=False,
+              help="Deploy Longhorn on k3s")
+@click.option("--longhorn-version", default=None,
+              help='Longhorn version to install (defaults to latest stable)')
```
kshtsk, line 967: Add a hint to the user about version format it consume, for example, it is not obvious that version has 'v' prefix.

kshtsk (changes requested):
Move the code of getting longhorn version to the python.

seslib/templates/k3s/provision.sh.j2
m-ildefons: I suggest using `helm` to install LH. It defaults to the latest stable release version but you can easily specify a specific version with the `--version` argument. No API queries or parsing or JSON manipulation needed.

That being said, I don't see this method as a hard blocker, since it gets the job done and the edge cases where it fails have little impact and aren't expected to happen a lot, given the user base and use cases of `sesdev` in the first place.

seslib/templates/k3s/provision.sh.j2
```diff
@@ -115,6 +121,57 @@ kubectl create -f rook-ceph/examples/toolbox.yaml
...
+  - http:
+      paths:
+      - pathType: Prefix
+        path: "/"
+        backend:
+          service:
+            name: longhorn-frontend
+            port:
+              number: 80
+" | kubectl -n longhorn-system create -f -
```
kshtsk, line 169: I thought we program sesdev in python, why do we do shell here at all?

seslib/templates/k3s/provision.sh.j2
kshtsk: > I suggest using `helm` to install LH. It defaults to the latest stable release version but you can easily specify a specific version with the `--version` argument. No API queries or parsing or JSON manipulation needed.

I like the idea to use helm for it.

> That being said, I don't see this method as a hard blocker, since it gets the job done and the edge cases where it fails have little impact and aren't expected to happen a lot, given the user base and use cases of `sesdev` in the first place.

Though, if we can avoid this now, why not to drop this ugly code? It costs nothing in comparison what we need to do in the future with it, going through debugging and trying to recall why is it happening wasting more time.

sesdev/__init__.py
tserong: I've gotten rid of the SP3 override, so it will now just default to Tumbleweed. I'll open a separate PR for adding SP4 and/or SP5 support (that's a general change anyway, not specific to this Longhorn work).

seslib/templates/k3s/provision.sh.j2
tserong: > another reason, why not to do this in shell script, if you are running it on multiple nodes, sequentially it may pick the inconsistent version from git hub, when there is new version appears in the middle

That piece of code only ever runs once on the master node so there's no risk of version drift.

sesdev/__init__.py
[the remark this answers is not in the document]
tserong: Right, I've gotten rid of that weird little bit of magic now :-)

seslib/templates/k3s/provision.sh.j2
tserong: That said, thanks guys - using helm does seem neater under the circumstances. Let me try that instead...

seslib/templates/k3s/provision.sh.j2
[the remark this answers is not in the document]
tserong: Immediately, I've added an extra guard to bail out if `/dev/vdb` doesn't exist.

> I think with longhorn you're supposed to pack all disks that you want LH to use into a LVM and then mount that. 

Yeah, I saw that in the docs but there's also the ability to have longhorn use multiple devices that are on separate mountpoints without LVM, and I wasn't sure which was the most recommended choice, so went with a single device initially just to get something running.

sesdev/__init__.py
tserong: Done

seslib/templates/k3s/provision.sh.j2
tserong: That's how sesdev does provisioning for each node. sesdev (written in python) generates a shell script, which is run on each node. The provisioning shell script includes all the commands needed to install and deploy everything.  See https://github.com/SUSE/sesdev/blob/master/seslib/templates/provision.sh.j2, which in turn includes various other scripts from under seslib/templates.

PR #693, opened 2023-04-21
k3s/longhorn: use LVM to support additional storage disks

tserong:
This is a really simple use of LVM to just create a linear logical volume over `/dev/vd[b-z]`, then mount that for use by Longhorn.  The default is still to deploy with only one additional disk, but at least now if more than one disk is requested, we'll actually do something with the extra disks.

### DIFF
--- a/seslib/templates/k3s/provision.sh.j2
+++ b/seslib/templates/k3s/provision.sh.j2
@@ -8,7 +8,7 @@ zypper --non-interactive install curl openssl
 export PATH=$PATH:/usr/local/bin
 
 {% if k3s_deploy_longhorn %}
-zypper --non-interactive install nfs-client open-iscsi e2fsprogs
+zypper --non-interactive install nfs-client open-iscsi e2fsprogs lvm2
 systemctl enable iscsid
 systemctl start iscsid
 {% endif %} {# k3s_deploy_longhorn #}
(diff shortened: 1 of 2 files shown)


tserong:
Hang on, just tried a deploy on Tumbleweed which failed for some reason...

tserong:
```
node4:~ # lvcreate --name longhorn-data -l 100%VG longhorn
  WARNING: failed to init udev for external device info, using none.
  /dev/longhorn/longhorn-data: not found: device not cleared
  Aborting. Failed to wipe start of new LV.
```

Huh?

tserong:
Deployed again, didn't have the problem >sigh<

I assume this is something potentially glitchy at some lower level than sesdev, so removing DNM label.

tserong:
Thinking about this further, I actually wonder if I should make this LVM setup optional, or perhaps just not bother doing it at all.

My reason for adding longhorn support to sesdev in the first place was so I'd have a convenient playground for experimenting with longhorn. If I stitch all the extra disks together with LVM automatically, that means I _can't_ play with Longhorn's "add disk" functionality...

seslib/templates/k3s/provision.sh.j2
```diff
@@ -193,12 +193,16 @@ curl -sfL https://get.k3s.io | K3S_URL=https://{{ master.fqdn }}:6443 sh -
...
 {% if k3s_deploy_longhorn %}
 if [ ! -b /dev/vdb ]; then
-    echo "ERROR: Longhorn deployments require one additional disk"
+    echo "ERROR: Longhorn deployments require at least one additional disk"
     false
 fi
-mkfs.ext4 /dev/vdb
+
+# This won't handle more than 25 disks.  Tim is pretty sure that's fine.
+vgcreate longhorn /dev/vd[b-z]
```
tserong, line 201: I probably _shouldn't_ call the vg `longhorn`, because that means we end up with `/dev/longhorn/longhorn-data`, and `/dev/longhorn` is the namespace used for _longhorn_ volumes.

PR #695, opened 2023-05-03
Jenkinsfile.integration: collect artifacts

kshtsk:


tserong (approved):
This looks neat!

Once this is merged I can always open another test PR which deliberately injects a failure earlier, so we can see what happens. I assume it should work fine though, as the cleanup just runs at the end.

PR #696, opened 2023-05-05
Inject a failure to make sure artifact collection works

tserong:


tserong:
Alright, let's see what Jenkins does with this

tserong:
LOL! This didn't even get to the point of hitting the fake failure I injected - instead, it seems to have failed earlier with that same weird intermittent failure that started this whole damn investigation. The important thing though, is that the log collection from #695 worked perfectly - thanks @kshtsk :-) Now I get to read the logs and see if I can figure out WTF went wrong...

PR #701, opened 2023-06-21
Jenkinsfile.integration: use `sesdev create --salt`

tserong:
After looking through ceph-salt logs from previous failed Jenkins runs, and also at Volker's issue which seems to be the same thing (see https://github.com/SUSE/sesdev/issues/700), my current suspicion is that the ceph-salt executor is correctly starting `salt -G 'ceph-salt:member' state.apply ceph-salt` but is then failing to pick up some (or all) of the event notifications, which results in it returning too soon, while `cephadm bootstrap` is still running.

Assuming I'm on the right track here, let's ask `sesdev` to run that salt command directly, to remove the ceph-salt executor from the picture entirely.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -68,7 +68,7 @@ def cloud_params = [
 ]
 
 def sesdev_create_cmd(ceph_salt_git_repo, ceph_salt_git_branch, extra_repo_urls) {
-    cmd = "sesdev create pacific --non-interactive"
+    cmd = "sesdev create pacific --salt --non-interactive"
     if (!ceph_salt_git_repo.isEmpty()) {
         if (ceph_salt_git_branch.isEmpty()) {
             error('ceph-salt git repo set, but no ceph-salt git branch given')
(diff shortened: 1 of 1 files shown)


kshtsk (approved):
looks trivial

tserong:
OK, did _not_ expect _that_ to fail :-/

tserong:
The first CI run (http://see.prv.suse.net:8080/blue/organizations/jenkins/sesdev-integration/detail/PR-701/1/ failed with:

```
    master: ++ salt -G ceph-salt:member state.apply ceph-salt
    master: master.mini.test:
    master:     Minion did not return. [No response]
    master:     The minions may not have all finished running and any remaining minions will return upon completion. To look up the return data for this job later, run the following command:
    master: 
    master:     salt-run jobs.lookup_jid 20230621113458675257
    master: ERROR: Minions returned with non-zero exit code
```

So then I added ded1bde625b5acc63022d64e93403992f8fe7bbb to collect more artifacts, but of course now everything is fine again - http://see.prv.suse.net:8080/blue/organizations/jenkins/sesdev-integration/detail/PR-701/4/ succeeded.

There's obviously still some intermittent problem :-/ but now we'll have even more stuff for me to look at next time it fails.

kshtsk (approved):
bless you

PR #704, opened 2023-06-26
Jenkinsfile.integration: use python3.11 for jcs

kshtsk:
It appears in the newer openstackclient code the new feature available starting from Python 3.7:
```
  from __future__ import annotations
```
For more info see:
  https://peps.python.org/pep-0563/#enabling-the-future-behavior-in-python-3-7

Since the default Python in Leap 15.4 is still 3.6, so maybe just start using the latest available 3.11 for jcs.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -143,10 +143,10 @@ pipeline {
                 }
                 sh """
                     mkdir -p ${compute_workspace}
-                    virtualenv ${jcs_venv}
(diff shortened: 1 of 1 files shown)


tserong (approved):

tserong:
> Something in test. I recommend to merge this PR it does not touch the code where the test failing.

Yeah, that's that irritating intermittent problem I've been trying to hunt down, and is unrelated to this fix.

Thanks!

PR #709, opened 2024-04-01
add quincy and reef support on ubuntu-focal

trociny:


tserong (approved):

tserong:
(I don't think that Jenkins CI build is ever going to happen)

